PR #4, opened 2017-05-08
fixes panic when terminal was resized

wose:
Some examples panicked when the terminal was resized. The canvas example also panicked when the *ball* hit the border of the canvas.


fdehau:
I knew some of the examples had this problem but I've never taken the time to investigate it. Thanks for the work.

PR #20, opened 2017-10-28
add note to readme about running/quitting examples

willbush:
This is a small change, but it was a bit confusing for me when I was evaluating if this repository could do what I want. I did not realize cargo could run examples until now, and I also couldn't figure out how to quit the example while it was running (I was trying things like Ctrl-C).


fdehau:
Thanks. The library definitively needs something in the form of a starting guide for people exploring it for the first time. I just need to find some time to write it properly :).

PR #30, opened 2017-12-21
Add Copy trait for layout enums

emsal1863:
Noticed that I need to call `.clone()` when I try to do something like pass in a variable of type `Direction` into a function call. Hope this is reasonable.


fdehau:
This is definitively a reasonable change ;). Thanks for pointing that out

Issue #40, opened 2018-03-19
Wrong operator in Buffer::pos_of?


Mange:
I might be misunderstanding this whole thing, but it looks like the wrong comparison operator is used in `Buffer::pos_of`.

https://github.com/fdehau/tui-rs/blob/c56173462ab9dd6a640d1811830a52f79f5f7a82/src/buffer.rs#L170-L176

The index must be *bigger* than the buffer size instead of *within* the buffer size.

Given a buffer of 20*20 cells (length 400), I would expect 0 to be (0, 0), 20 to be (0, 1), and so on. Instead I must pass an index like 400 or 555.

PR #41, opened 2018-03-22
BUG: Buffer::pos_of panics on inside-bounds index

Mange:
- Add tests for this behavior.
- Extend documentation of Buffer::pos_of and Buffer::index_of
  - Clarify that the coordinates should be in global coordinate-space (rather than local).
  - Document panics.
  - Add examples.

Adding both tests and doctests for these conditions might be overkill, but I hope there might be value in having both. Tests can be extended with more edge cases later without muddying up the documentation, for example.

Please let me know if you think this is appropriate or not.

Fixes #4.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -153,7 +153,34 @@ impl Buffer {
         &mut self.content[i]
     }
 
-    /// Returns the index in the Vec<Cell> for the given (x, y)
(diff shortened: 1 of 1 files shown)


fdehau:
Thanks you for this PR, this neat contribution is very much appreciated. If I'm not mistaken this also fixes #40.

Mange:
Whoa, so think #4 was a typo. I don't know if this would fix that issue. Sorry!

PR #48, opened 2018-04-03
Fix colors use, add missing Blue

bestouff:
Colors use is incorrect:
- Obviously `Blue` was missing.
- There was a mixup between `Gray`, `DarkGray` and `White`, and some of them had been hardcoded to 24-bits rgb values instead of using the terminal's values.

This commit fixes both.


fdehau:
Thanks =)

PR #55, opened 2018-05-14
Text alignment option for paragraph widget

man0lis:
I needed this functionality for a project I'm working on and thought it made sense to implement it in the widget itself.
The multipeek implementation is taken from itertools to avoid pulling it as a dependency.
Word wrap seems to work.
I have not tested this implementation with fancy unicode characters.

Failed travis is due to formatting errors in other files, I did not touch.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ default = ["termion"]
 [dependencies]
 bitflags = "1.0.1"
 cassowary = "0.3.0"
+itertools = "0.7.8"
 log = "0.4.1"
 unicode-segmentation = "1.2.0"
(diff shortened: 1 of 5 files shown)


fdehau:
Thank you for this contribution. That looks great. While I understand your concern regarding pulling another dependency I would rather have itertools as a dependency to not have to maintain code not directly related to this crate. In addition, the library is already using iterators in many places and there is a chance that some patterns may be improved using this crate. If you could do that I'm happy to merge =)

man0lis:
Thank you for considering my contribution :)
I removed the multipeek code, added itertools as a dependency and rebased the branch on master.

Issue #70, opened 2018-09-05
Feedback on v0.3.0-beta.0


abusch:
I tried to migrate one of my projects to the newly released v0.3.0-beta.0. For the most part things look good. The one thing that I had a problem with are the changes to `Paragraph`, which now takes of a `Vec<Text>`, and the fact that `Text` only takes borrowed strings. It made it really hard to do things like:

```rust
let mut text = vec![];
if some_condition {
    text.push(&format!("..."));
} else {
    text.push(&format!("..."));
}
Paragraph::new(text.iter())...
```
due to lifetime issues (strings returned by `format!()` don't live long enough). I think maybe `Text` should store a `Cow<str>` internally, so it could accept both borrowed and owned strings.

What do you think?

I might give it a shot on a branch an see how it goes...

PR #71, opened 2018-09-06
Make Text accept both borrowed and owned strings

abusch:
I took a crack at fixing issue #70 and here's what it looks like. The slightly annoying thing is having to add `.into()` when building a `Text`. If that's too annoying we can maybe add some helper methods like `Text::data()` and `Text::styled_data()` that would do the conversion for you.

### DIFF
--- a/examples/demo.rs
+++ b/examples/demo.rs
@@ -439,21 +439,21 @@ fn draw_charts(f: &mut Frame<MouseBackend>, app: &App, area: Rect) {
 
 fn draw_text(f: &mut Frame<MouseBackend>, area: Rect) {
     let text = [
-        Text::Data("This is a paragraph with several lines. You can change style your text the way you want.\n\nFox example: ".into()),
-        Text::StyledData("under".into(), Style::default().fg(Color::Red)),
-        Text::Data(" ".into()),
-        Text::StyledData("the".into(), Style::default().fg(Color::Green)),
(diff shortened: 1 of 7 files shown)


abusch:
I added some helper methods `Text::data()` and `Text::styled_data()` to make building `Text` instances a bit easier (i.e without having to call `into()`). Let me know what you think!

Note that if we go with this, it might be worth changing `Item` the same way for consistency (right now it takes an owned string, and reference to a style, which is the exact opposite of `Text`...)

fdehau:
Hi, thanks for taking the time to look into this. I agree with most of the changes.

Item currently requires its data to implements `Display`. It thus works for both owned or borrowed strings but all items should be of the same type. So `Cow` might also be a better idea there, especially when items are created "manually".  As of the `Style`, I think that we should go for an owned version.

An unified definition of `Text` and `Item` could be:

```rust
pub enum Text<'b> {
    Raw(Cow<'b, str>),
    Styled(Cow<'b, str>, Style),
}
```
dropping `Data` in names which appears unnecessary the more I read it.

Let me know what you think...

abusch:
Yes, I think that makes sense. I've implemented this so you can see what it looks like.

fdehau:
Great work :+1: Let me know if you have more feedback on the upcoming release.

PR #95, opened 2018-11-27
 Gauge: use f64 internally and allow to set any f64 between 0 and 1 

karolinepauls:
I wanted to also change `Constraint` but it's a bit more complex (needs a Rational or a f64 with Hash and Eq), another time.

### DIFF
--- a/src/widgets/gauge.rs
+++ b/src/widgets/gauge.rs
@@ -22,7 +22,7 @@ use widgets::{Block, Widget};
 /// ```
 pub struct Gauge<'a> {
     block: Option<Block<'a>>,
-    percent: u16,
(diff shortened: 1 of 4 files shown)


src/widgets/gauge.rs
```diff
@@ -104,3 +118,21 @@ impl<'a> Widget for Gauge<'a> {
         }
     }
 }
+
+#[test]
```
fdehau: Shouldn't those be put inside a `tests` module (as described here: https://doc.rust-lang.org/stable/book/2018-edition/ch11-03-test-organization.html) ? I thought that it was required for them not to be included in a normal build.

PR #97, opened 2018-12-02
Make sure we always emit a cursor goto for the first update.

deifactor:
Otherwise, if the first update is to (1, 0) then no goto occurs. Something like this reproduces the error, producing "Playause" on the second update.

```rust
use termion::raw::IntoRawMode;
use tui;
use tui::widgets::Widget;

fn main() -> Result<(), failure::Error> {
    let stdout = std::io::stdout().into_raw_mode()?;
    let stdout = termion::screen::AlternateScreen::from(stdout);
    let backend = tui::backend::TermionBackend::new(stdout);
    let mut terminal = tui::Terminal::new(backend)?;
    let size = terminal.size()?;

    for i in ["Play", "Pause"].iter() {
        terminal.draw(|mut f| {
            let line = [tui::widgets::Text::raw(i.clone())];
            tui::widgets::Paragraph::new(line.iter()).render(&mut f, size);
        });
        std::thread::sleep(std::time::Duration::from_millis(1000));
    }
    Ok(())
}
```

### DIFF
--- a/src/backend/crossterm.rs
+++ b/src/backend/crossterm.rs
@@ -59,9 +59,11 @@ impl Backend for CrosstermBackend {
         let cursor = crossterm::cursor();
         let mut last_y = 0;
         let mut last_x = 0;
+        let mut first = true;
         for (x, y, cell) in content {
-            if y != last_y || x != last_x + 1 {
+            if y != last_y || x != last_x + 1 || first {
                 cursor.goto(x, y);
+                first = false;
             }
             last_x = x;
             last_y = y;
--- a/src/backend/termion.rs
+++ b/src/backend/termion.rs
@@ -70,7 +70,7 @@ where
         let mut last_x = 0;
         let mut inst = 0;
         for (x, y, cell) in content {
-            if y != last_y || x != last_x + 1 {
+            if y != last_y || x != last_x + 1 || inst == 0 {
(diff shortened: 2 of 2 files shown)


src/backend/crossterm.rs
```diff
@@ -59,9 +59,11 @@ impl Backend for CrosstermBackend {
         let cursor = crossterm::cursor();
         let mut last_y = 0;
         let mut last_x = 0;
+        let mut first = true;
         for (x, y, cell) in content {
-            if y != last_y || x != last_x + 1 {
+            if y != last_y || x != last_x + 1 || first {
```
fdehau: Rather than evaluate this condition for all iterations of the loop, wouldn't it be better to explicitly check for this case beforehand. Using something like `peek` on the iterator ?

```rust
        let mut content = content.peekable();
        if let Some((x, y, _)) = content.peek() {
            if *x == 1 && *y == 0 {
                string.push_str(&format!("{}", termion::cursor::Goto(x + 1, y + 1)));
                inst += 1;
            }
        }

```

fdehau: I didn't know about the internal branching of `Peekable::next`, thanks for pointing that out =). Then you make a totally valid point and I will merge as is.

Issue #99, opened 2018-12-07
Auto resizing causes a race condition


dotdash:
The auto-resizing implemented in #94 unfortunately makes it impossible to ensure that you're using the right size in your layout calculation.

Previously you'd use something like:
```rust
let new_size = terminal.size()?;
if size != new_size {
    size = new_size;
    terminal.resize(size)?;
}

terminal.draw(|mut f| { /* ... */ });
```
which ensures that the `size` variable later used to calculate the layout is the same size that is used for the buffer inside the `terminal` instance.

But now you have to do:
```rust
let size = terminal.size();
/* RACE WINDOW HERE */
terminal.draw(|mut f| { /* ... */ });
})
```

You have to ask for the size of the terminal first, and then, inside the draw() method, the size might change because of the automatic resizing. If the terminal is actually resized in between these two actions, the sizes get out of sync, and there's no way for the programmer to avoid the resulting artifacts or crash.

One way to handle this would be to adjust the callback in `draw()` to accept the current terminal size as a second parameter like this:

```rust
terminal.draw(|mut f, size| { /* ... */ });
```

That would ensure that the size matches the buffer size. If that sounds good to you, I could prepare a PR for that.

dotdash:
cc @karolinepauls

karolinepauls:
Another way is to turn Frame into a "rendering transaction object", with a consistent Terminal state snapshot as of render time. Then the "render-time consistent" size can be obtained from the Frame.

PR #100, opened 2018-12-07
Frame: provide consistent size for rendering

karolinepauls:
That's one way of handling #99 . Please review.

cc @dotdash

### DIFF
--- a/examples/barchart.rs
+++ b/examples/barchart.rs
@@ -77,13 +77,12 @@ fn main() -> Result<(), failure::Error> {
     let mut app = App::new();
 
     loop {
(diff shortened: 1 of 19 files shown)


karolinepauls:
The new stable toolchain shipped a new rustfmt. Formatting fixed in: https://github.com/fdehau/tui-rs/pull/101.

When pulls/101 gets merged, let me fix the conflicts.

fdehau:
Thanks for taking the time to quickly fix this. Could you also update the doc examples in lib.rs to get newcomers to use this approach ?

PR #103, opened 2018-12-10
Paragraph - line wrapping on word boundaries

karolinepauls:
There's a bunch of things to improve and test cases to add. LineWrapper is also ugly but I'll be making it prettier when I have tests.

The LineComposer trait implementations cannot be iterators because we attempt to be efficient so we only yield slices of the current line which lives only one "iteration".

The changeset doesn't modify the public API, however the behaviour of `.wrap(true)` has been changed to do wrapping on word boundaries. Since I used runtime polymorphism, we could allow users to inject their own wrapping logic but the interface is ugly (consumes and passes Style), so I'd keep it private.

We may want to reintroduce opt-in previous behaviour of wrapping regardless of previous behaviour but this would require to make `.wrap` take an enum.

### DIFF
--- a/src/widgets/reflow.rs
+++ b/src/widgets/reflow.rs
@@ -1,0 +1,446 @@
+use style::Style;
+use unicode_width::UnicodeWidthStr;
+
+const NBSP: &str = "\u{00a0}";
+
+#[derive(Copy, Clone, Debug)]
+pub struct Styled<'a>(pub &'a str, pub Style);
+
+/// A state machine to pack styled symbols into lines.
+/// Cannot implement it as Iterator since it yields slices of the internal buffer (need streaming
+/// iterators for that).
+pub trait LineComposer<'a> {
+    fn next_line(&mut self) -> Option<(&[Styled<'a>], u16)>;
+}
+
+/// A state machine that wraps lines on word boundaries.
+pub struct WordWrapper<'a, 'b> {
+    symbols: &'b mut Iterator<Item = Styled<'a>>,
+    max_line_width: u16,
+    current_line: Vec<Styled<'a>>,
+    next_line: Vec<Styled<'a>>,
+}
+
+impl<'a, 'b> WordWrapper<'a, 'b> {
+    pub fn new(
+        symbols: &'b mut Iterator<Item = Styled<'a>>,
+        max_line_width: u16,
+    ) -> WordWrapper<'a, 'b> {
+        WordWrapper {
+            symbols,
+            max_line_width,
+            current_line: vec![],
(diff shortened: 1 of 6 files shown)


fdehau:
That looks very promising. I particularly enjoy the fact that the you were able to move some logic out of the `draw` method. Moreover, I think that we will be able to use those "line composers" in List too...

I don't mind that that the behavior of the `wrap` changes since I think it is what people usually expect from text wrapping. My first approach was the simplest solution I could think of with all the shortcomings that this implies ;).

With that, the changes requested to `List` and all the recent work from you and other contributors, I think we are good for a "0.4" release anyway.

karolinepauls:
So when writing tests I realised I had accidentally rewritten the old implementation of wrapping instead of line truncation. I fixed it by writing another composer but now we're left with the older one, usable but unused. We could actually expose it but it would require changing the public interface - `wrap` method.

I have to make text wrapping behave sanely when the width is very low - decide how to treat whitespace and whether to display double width characters even if the width is 1 (I don't think so - we should either skip all characters wider than the current width or we should refuse to render below 2 chars of width - but what if there are triple width characters?)

karolinepauls:
Removed the unused composer, added test cases, added skipping of characters wider than the text area, corner cases.

@fdehau Ready for review.

src/widgets/reflow.rs
```diff
@@ -0,0 +1,446 @@
...
+    fn line_composer_char_plus_lots_of_spaces() {
+        let width = 20;
+        let text = "a                                                                     ";
+        let (word_wrapper, _) = run_composer(Composer::WordWrapper, text, width);
+        let (line_truncator, _) = run_composer(Composer::LineTruncator, text, width);
+        // What's happening below is: the first line gets consumed, trailing spaces discarded,
+        // after 20 of which a word break occurs (probably shouldn't). The second line break
+        // discards all whitespace. The result should probably be vec!["a"] but it doesn't matter
+        // that much.
+        assert_eq!(word_wrapper, vec!["a", ""]);
```
karolinepauls: I'm not the fan of this but handling it would require to make the composer a bit more high level.

Since the case isn't critically important (and will never be visible when rendered), I don't want to make this PR dependent on that.

Improving the composer should go through another PR.

fdehau: This is totally fine for a first version. As you said this should not matter to end users and you documented the edge case with this test.

fdehau:
I finally had some time to review your proposed changes. This is well thought and properly tested. Going through the code, I've not found anything to fix. Thus thank you for this work and looking forward for other contributions of this quality ;)

PR #105, opened 2018-12-14
Buffer: double width handling

karolinepauls:
Waits for: https://github.com/fdehau/tui-rs/pull/103

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -3,6 +3,7 @@ use std::fmt;
 use std::usize;
 
 use unicode_segmentation::UnicodeSegmentation;
+use unicode_width::UnicodeWidthStr;
 
 use layout::Rect;
 use style::{Color, Modifier, Style};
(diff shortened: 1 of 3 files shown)


src/buffer.rs
```diff
@@ -362,12 +389,72 @@ impl Buffer {
...
+    /// Updates: `0: コ` (double width symbol at index 0 - skip index 1)
+    /// ```
+    ///
+    /// ```text
+    /// (Index:) `012`
+    /// Prev:    `aaa`
+    /// Next:    `aコ`
+    /// Updates: `0: a, 1: コ` (double width symbol at index 1 - skip index 2)
+    /// ```
+    pub fn diff<'a>(&self, other: &'a Buffer) -> Vec<(u16, u16, &'a Cell)> {
```
karolinepauls: The previous algorithm would be enough if we could guarantee that all phantom cells (extra cells occupied by a multi-width char) are empty (`symbol: ""`), while regular empty cells are blank (`symbol: " "`), which would make them different. This would require care in every place text is rendered - always using either `set_string`, `set_stringn`, or some other method which takes care of variable symbol width, never inserting a symbol manually with `buf.get_mut().set_symbol()`.

Right now Paragraph happens to produce a well-formed buffer despite (`buf.get_mut().set_symbol()`) because the background is always redrawn.

svenstaro:
You need to rebase this.

PR #106, opened 2018-12-16
[crossterm] Remove #attr() styling call when compiling on non-unix systems.

Jezza:
The #attr() isn't available on unix systems, and so it fails to compile on windows.

This is pretty difficult to fix upstream, as windows doesn't really support a whole lot of attributes.

A good idea might be to think of this as a quick fix, and I'll open an issue on crossterm, as it does support _some_ attributes, just not many.
So a better solution should be devised, but until then, we should probably just drop attributes.

https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ Cargo.lock
 *.log
 *.rs.rustfmt
 .gdb_history
+/.idea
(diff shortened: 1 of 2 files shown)


svenstaro:
You need to rebase this.

Jezza:
Actually, good news is this PR can be closed.
I think the latest version of crossterm supports it.

So, just need to bump the crossterm version, I think.

svenstaro:
Well then maybe just make a PR with that.

fdehau:
This was solved by @Jezza in https://github.com/TimonPost/crossterm/issues/60. The next release will have the last version of the `crossterm` with this fix.

PR #109, opened 2019-01-06
Add new shape: Rectangle

### LINKED ISSUE #87: Add Box shape
I think there should be a Box shape just like there is a Line shape.

svenstaro:
This currently depends on #106 and fixes #87.

Renaming `Line` to `LineShape` is for consistency with `BoxShape` as I couldn't just call it `Box` and `Rect` is already in use. Also, there are already other uses of `Line` in this library so it would be helpful to disambiguate those too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # Changelog
 
-## v0.3.1 - To be released
+##  To be released
 
+* Added a new shape: Rectangle
 * Upgraded to Rust 2018 edition
 
 ## v0.3.0 - 2018-11-04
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 # Changelog
 
-## To be released
+## v0.4.0 - To be released
```
fdehau: I appreciate the effort of keeping the changelog up to date but I would rather have everything inside "to be released". I'm not certain that I will follow this release steps (especially the rust 2018 edition in a patch version)

Cargo.toml
```diff
@@ -20,7 +21,7 @@ default = ["termion"]
 [dependencies]
 bitflags = "1.0"
 cassowary = "0.3"
-itertools = "0.7"
+itertools = "0.8"
```
fdehau: Is there a reason for this update to be part of this PR ? I try to update all dependencies before each major release and this require some testing to be sure that no strange behaviors are introduced.

svenstaro:
@fdehau There you go.

svenstaro:
Rebased

PR #121, opened 2019-01-24
Add: pancurses backend

defiori:
On Unix I don't see a reason not to use termion, but on Windows [pancurses](https://github.com/ihalila/pancurses) adds the [PDcurses](https://github.com/Bill-Gray/PDCurses) backend which works quite nicely. (Plus you get Unix ncurses for free, however without Unicode).


defiori:
I understand your hesitation to keep adding backends, however I think pancurses would be worthwhile. For me two reasons stand out:

- **Windows**: I initially added the pancurses backend to make tui usable on Windows. I'm aware that [crossterm](https://github.com/TimonPost/crossterm) aims to be a full cross-platform terminal library and truly hope they become the go-to solution in Rust. However, their API is still [in flux](https://github.com/TimonPost/crossterm/blob/master/docs/UpgradeManual.md) and there are a few significant feature gaps, e.g. keyboard input (see [this](https://github.com/TimonPost/crossterm/issues/81) and [this](https://github.com/TimonPost/crossterm/issues/80)). While I'm aware that tui doesn't aim to deal with input itself, it should still be able to expose reasonable input handlers from the backend.

- **Curses**: I think it's fair to say that *curses isn't just yet another terminal library but standard for text-based applications (htop, ranger, ..). While that's not an argument in itself, it means that the API is set in stone and there shouldn't be any stability issues down the road. On *nix `pancurses` is quite a thin wrapper around the system ncurses libraries which some users may prefer to bundling a full terminal library. On Windows it comes with a full terminal implementation (PDcurses win32a flavor) that talks directly to the Windows API, making it fully cross-platform.

In summary, while I know the crossterm devs are working hard on filling in the gaps (e.g. [input handling](https://github.com/TimonPost/crossterm/issues/81)) the pancurses backend already offers full Windows functionality plus a few goodies that will be useful even as crossterm gets more mature.

src/backend/curses.rs
```diff
@@ -0,0 +1,231 @@
...
+#[cfg(unix)]
+use unicode_segmentation::UnicodeSegmentation;
+
+pub struct CursesBackend {
+    curses: easycurses::EasyCurses,
+}
+
+impl CursesBackend {
+    pub fn new() -> Result<CursesBackend, String> {
+        match easycurses::EasyCurses::initialize_system() {
```
fdehau: Unless part of this setup is required for this backend to work I would rather have the user give the constructor a configured curses instance instead of doing the configuration internally. I have already experimented with this in the Termion backend and I have found that it is very difficult to provide a configuration that match all use cases.

defiori: That makes sense, I've kept `new()` as a convenience constructor and added a `with_curses()` method that takes a configured instance. This is in line with how the rustbox and crossterm backends handle things.

defiori:
I just pushed the changes, curses_demo now uses the shared ui / app code and I've removed the standalone curses example. See comment on change request for changes to the backend instantiation.

fdehau:
LGTM, thanks for taking the time to work on this and for being patient enough with my delayed reviews =)

PR #125, opened 2019-02-02
Create Crossterm Backend with Previously Created Screen

JonBoyleCoding:
Added the ability to create the crossterm backend with a previously created screen. This will allow the API user to specify whether they want to enter into raw mode and perform any prior setup on the crossterm end before passing off to tui.

### DIFF
--- a/src/backend/crossterm.rs
+++ b/src/backend/crossterm.rs
@@ -16,6 +16,10 @@ impl CrosstermBackend {
         }
     }
 
+    pub fn with_crossterm(scr: crossterm::Screen) -> CrosstermBackend {
+        CrosstermBackend { screen: scr }
+    }
+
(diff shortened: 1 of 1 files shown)


src/backend/crossterm.rs
```diff
@@ -16,6 +16,10 @@ impl CrosstermBackend {
         }
     }
 
+    pub fn with_crossterm(scr: crossterm::Screen) -> CrosstermBackend {
```
fdehau: To be a bit more consistent with the actual parameter, what about:
```suggestion
    pub fn with_screen(screen: crossterm::Screen) -> CrosstermBackend {
```

JonBoyleCoding: Sounds good.

JonBoyleCoding:
I've made the change suggested by @fdehau.

PR #128, opened 2019-02-08
Add alternate screen mode to crossterm backend

defiori:
`CrosstermBackend` currently does not allow alternate screen mode where we return to the original screen once we're done. #125 doesn't seem to fix this because we can't move a screen out of [AlternateScreen](https://docs.rs/crossterm/0.6.0/crossterm/struct.AlternateScreen.html) because of `Drop` (see [this](https://github.com/TimonPost/crossterm/issues/74) crossterm issue). This PR adds the required methods to `CrosstermBackend` and adds `crossterm_demo.rs` as an example.

### DIFF
--- a/src/backend/crossterm.rs
+++ b/src/backend/crossterm.rs
@@ -7,7 +7,7 @@ use crate::style::{Color, Modifier};
 use crossterm::error::ErrorKind;
 
 pub struct CrosstermBackend {
-    screen: crossterm::Screen,
+    screen: Option<crossterm::Screen>,
     crossterm: crossterm::Crossterm,
     // Need to keep the AlternateScreen around even when not using it directly,
     // see https://github.com/TimonPost/crossterm/issues/88
@@ -19,7 +19,7 @@ impl Default for CrosstermBackend {
         let screen = crossterm::Screen::default();
         let crossterm = crossterm::Crossterm::from_screen(&screen);
         CrosstermBackend {
-            screen,
+            screen: Some(screen),
             crossterm,
             alternate_screen: None,
         }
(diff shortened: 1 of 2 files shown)


src/backend/crossterm.rs
```diff
@@ -8,12 +8,20 @@ use crossterm::error::ErrorKind;
 
 pub struct CrosstermBackend {
     screen: crossterm::Screen,
+    crossterm: crossterm::Crossterm,
+    // Need to keep the AlternateScreen around even when not using it directly,
+    // see https://github.com/TimonPost/crossterm/issues/88
```
fdehau: Even though I understand the issue, I doubt that the tui crate should cope explicitly with this issue. I would rather have the user aware of the limitations or specific behaviors of the backend she/he is using. Wouldn't a comment in the documentation be enough ?

Cargo.toml
```diff
@@ -46,3 +46,8 @@ required-features = ["rustbox"]
 name = "crossterm"
 path = "examples/crossterm.rs"
 required-features = ["crossterm"]
+
+[[example]]
```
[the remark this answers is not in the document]
defiori: Updated in latest version.

src/backend/crossterm.rs
defiori: I just pushed a cleaned-up version which I think uses crossterm in the most coherent way, where all `Backend` methods only interact with the `crossterm::Crossterm` instance. This seems like a good idea to me independently of the alternate screen issue. 

Concerning the alternate screen: I agree that keeping the `AlternateScreen` instance around is something crossterm should ultimately deal with (I filed an issue [here](https://github.com/TimonPost/crossterm/issues/88)). However, before this is fixed on the crossterm side holding on to the `AlternateScreen` seems like a small price to pay for a working alternate screen mode. Alternatively, as you say the docs could mention that alternate screen mode is currently not implemented for the crossterm backend. It comes down to a trade-off between working around some issues in the backends and missing features.

fdehau: Ok presented that way you are right that it is a small price to pay. I will open an issue to not forget to revisit this problem if `crossterm` updates its behavior.

PR #133, opened 2019-02-25
Support several modifiers and indexed colors at once

### LINKED ISSUE #122: More than one modifier on a style
As the title says. Right now there's no way to make text that's (for example) both bold *and* italic. So either the `modifier` field on `Style` should be a set/vec/whatever or `Modifier` itself should be able to contain more than one individual modifier.

dflemstr:
This adds support for some more modifiers, several concurrent modifiers, and direct access to indexed colors.

Fixes #122.


fdehau:
Great work. From a first look, that's mostly what I had in mind so I don't expect to request big changes from you. Since this is a breaking change in the public api, I will leave this open for a few days to let people comment if they have mixed feeling about this. As rightfully highlighted by @defiori in the issue, performance should not be the first concern but since you have done most of the work already, have you by any chance profiled the code ? My main concern with the bitflags implementation was that it would introduce a lot of branching in one of the main hotpath of the library and I was hoping that the compiler would optimized them out.

dflemstr:
@fdehau I did several benchmarks, not microbenchmarks as I think they can be deceiving in this case, but instead sampling some real world applications I'm building (e.g. a terminal emulator running `unimatrix`) using Linux `perf` and `flamegraph`.

Here's the resulting flamegraph (sadly I can't embed the actual SVG since GitHub doesn't allow it):

image

As you can see, most of the time in a `tui` `draw` call is currently spent on the `flush` (58.0% of `draw`), and the `diff` call takes most of that time (93.8% of `flush`)

If you actually look at the Termion draw backend, almost all of the time is spent on `std::io::Write::write_fmt` since the code is allocating a `String`.

In summary, if you want to optimize this code, there are a lot more important bottle necks due to memory allocation and random access memory compared to branching. I would start by:

  1. Optimizing `diff`, which will have a 16 times higher performance impact than spending time on the drawing backends.  Specifically, most of the time is spent in the `unicode_width` crate, maybe some calls to that can be avoided?
  2. Optimizing `tui::buffer::Buffer::reset` and `tui::widgets::block::Block::draw` since they set a lot of memory to the same value(s) in a way that thrashes the CPU memory cache right now.
  3. If you really want to optimize the terminal backends, first remove the allocations of strings/vecs etc.
  4. The particular branches needed for `Modifier` only constitutes 0.15% of runtime (not statistically significant, I only received one sample) so I don't think it's worth it to spend any time there.

Hope this helps!

src/style.rs
```diff
@@ -33,62 +36,72 @@ impl Color {
             Color::Blue => "b",
             Color::Magenta => "m",
             Color::Cyan => "c",
-            Color::Gray => "g",
-            Color::DarkGray => "G",
+            Color::Gray => "w",
+            Color::DarkGray => "B",
```
dflemstr: For the record, I made these changes to aid debugging, since the codes were ambiguous before (`G` = light green and gray) and additionally "gray" is usually called "white", "dark gray" usually called "light black" and "white" usually called "light white" in other terminals.

fdehau: That is totally fine with me. You are right that the existing debug codes were not well thought in the first place.

fdehau:
So that was a totally misleaded assumption on my part =). Thank you for taking the time to make this thorough analysis. The good news is those changes are then ok for me. However, in the meantime I've merged a new backend based on pancurses and easycurses. Would you perform the necessary changes there too regarding the updated style struct for the build to pass with all features activated ?

dflemstr:
@fdehau yes, sounds reasonable, I will do that!

PR #134, opened 2019-02-25
Support exact ratios for layout constraints

dflemstr:


src/layout.rs
```diff
@@ -24,6 +24,7 @@ pub enum Direction {
 pub enum Constraint {
     // TODO: enforce range 0 - 100
     Percentage(u16),
+    Ratio(u32, u32),
```
fdehau: Is there a reason for not using an `f64` directly ?

dflemstr: Then this type can no longer derive `Eq` or `Hash`, which would break the current caching system for layout re-calculation.

PR #140, opened 2019-03-13
fix typo of CHANFELOG

longwusha:
fix the render of title in CHANFELOG


longwusha:
@fdehau the render result of title in CHANFELOG has some problem.

PR #141, opened 2019-03-13
 Add an example of image rendering

longwusha:


longwusha:
@fdehau Please help check the code, the rendering results are as follows:

深度截图_deepin-terminal_20190313224317

fdehau:
IMO examples are there to showcase some basic use cases of the widgets of this library. This example is both too advanced and a bit too far away from the common usage of this crate. So I'm afraid that I will not merge this. However, if you manage to put together a public project and you are still using `tui` I would happily link to it.

PR #150, opened 2019-04-12
Crossterm update 0.9.2

TimonPost:
Updated the version of crossterm, notice that it has API breaking changes. The 0.9.^ version has a lot of improvements.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ unicode-segmentation = "1.2"
 unicode-width = "0.1"
 termion = { version = "1.5", optional = true }
 rustbox = { version = "0.11", optional = true }
-crossterm = { version = "0.6", optional = true }
+crossterm = { version = "0.9.2", optional = true }
 easycurses = { version = "0.12.2", optional = true }
 pancurses = { version = "0.16.1", optional = true, features = ["win32a"] }
 
--- a/src/backend/crossterm.rs
+++ b/src/backend/crossterm.rs
@@ -135,6 +135,10 @@ impl Backend for CrosstermBackend {
         let mut last_y = 0;
         let mut last_x = 0;
         let mut first = true;
+
+        let stdout = io::stdout();
+        let mut handle = stdout.lock();
+
         for (x, y, cell) in content {
             if y != last_y || x != last_x + 1 || first {
                 cursor.goto(x, y).map_err(convert_error)?;
@@ -151,7 +155,7 @@ impl Backend for CrosstermBackend {
             }
             s.object_style.attrs = cell.style.modifier.into();
 
-            self.crossterm.paint(s).map_err(convert_error)?;
+           write!(handle, "{}", s);
         }
         Ok(())
     }
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ Cargo.lock
 *.log
 *.rs.rustfmt
 .gdb_history
+.idea/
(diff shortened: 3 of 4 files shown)


.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
fdehau: I would rather not have any editor specific config files checkout in the source code. Could you remove this ?

Cargo.toml
```diff
@@ -30,7 +31,7 @@ unicode-segmentation = "1.2"
 unicode-width = "0.1"
 termion = { version = "1.5", optional = true }
 rustbox = { version = "0.11", optional = true }
-crossterm = { version = "0.6", optional = true }
+crossterm = { optional = true, git = "https://github.com/TimonPost/crossterm", branch = "fixes" }
```
fdehau: I assume that it is meant to be `{ version = "0.9", optional = true }`

src/backend/crossterm.rs
```diff
@@ -151,7 +129,7 @@ impl Backend for CrosstermBackend {
             }
             s.object_style.attrs = cell.style.modifier.into();
 
-            self.crossterm.paint(s).map_err(convert_error)?;
+           write!(io::stdout(), "{}", s);
```
fdehau: Isn't this locking stdout every time you write ? Shouldn't you lock once and write to `StdoutLock` ?

Cargo.toml
TimonPost: No, currently I have to do a release with the changes on this branch, it has a couple of fixes, also useful for TUI. I think that release could be done this week when it is done, I will update this PR with the new version.

.idea/vcs.xml
TimonPost: Yes, I have to add .idea/ to the gitignore.

jonlil: A suggestion for you is to have folders like that in your personal gitignore ($HOME/.gitignore)

src/backend/crossterm.rs
TimonPost: I can try that, notice that that happened before too.

Cargo.toml
```diff
@@ -17,8 +17,9 @@ travis-ci = { repository = "fdehau/tui-rs" }
 appveyor = { repository = "fdehau/tui-rs" }
 
 [features]
-default = ["termion"]
+default = ["crossterm"]
```
TimonPost: can be changed when finished

src/backend/crossterm.rs
fdehau: I could misunderstand the [documentation](https://doc.rust-lang.org/std/io/fn.stdout.html) but for me locking outside the for loop would avoid the implicit synchronization. Furthermore, the struct returned by `io::stdout().lock()` is also writable.

Cargo.toml
fdehau: Yeah I would rather not change the default backend for now. `termion` has been stable for a longer period of time. I could reconsider this in the future when `crossterm` reaches a stable release.

src/backend/crossterm.rs
TimonPost: Right, I will take a look at if this could be done more efficiently.

Cargo.toml
TimonPost: agreed

src/backend/crossterm.rs
TimonPost: I moved the lock out of the loop, by doing this we only lock outside the loop, this lock is then used in the loop for drawing all cells.

TimonPost:
Woops, I messed up my git, I'll create an new PR later.

PR #154, opened 2019-05-17
Relaxing crossterm dependency version

svartalf:
This is a follow-up to #153.

Since `crossterm 0.9` version development is still in progress (as an example, see [this bug](https://github.com/TimonPost/crossterm/issues/141)), forcing to use one specific version is a little bit excessive, so this PR relaxes `crossterm` version to `^0.9`.
@TimonPost confirmed that `crossterm` adheres to semantic versioning, therefore version can be relaxed safely.

As an addition, I removed the `crossterm_backend` feature because of two reasons:
 1. `crossterm` feature is exists already (because `crossterm` as a depedency is optional)
 2. Code that introduces it was merged into master 6 hours ago (at the time of creating this PR) and was not published yet, therefore possibility of breaking something is really small.

P.S. I have personal request also - can there be new version of `tui` published after merge of this PR? This is a only thing which [blocks](https://github.com/svartalf/rust-battop/issues/5) me.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,6 @@ appveyor = { repository = "fdehau/tui-rs" }
 [features]
 default = ["termion"]
 curses = ["easycurses", "pancurses"]
-crossterm_backend = ["crossterm"]
 
 [dependencies]
 bitflags = "1.0"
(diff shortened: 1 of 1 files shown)


fdehau:
Merged =). I was about to revert the crossterm_backend feature myself so that made perfect sense to me. I've also published a new version (v0.6.0).

PR #155, opened 2019-05-19
Provide Clone and Debug for Text type

DarrienG:
I know using clone isn't the best practice, but in some cases it can be extremely convenient. I'm working on a project where I'm doing quite a dance to get around it. In the end I wondered why I wasn't using clone and the answer wasn't obvious.

Am currently using a local fork with this in the meantime.

### DIFF
--- a/src/widgets/mod.rs
+++ b/src/widgets/mod.rs
@@ -47,6 +47,7 @@ bitflags! {
     }
 }
 
+#[derive(Clone, Debug)]
 pub enum Text<'b> {
     Raw(Cow<'b, str>),
     Styled(Cow<'b, str>, Style),


alok:
Incidentally, public structs/enums should probably implement Debug, since it's usually handy. Deriving Eq is probably a good idea for text as well, since conceptually it's a string, and should have similar traits.

DarrienG:
Since we could debate about how to implement Eq (does it compare styles or only text? does it compare backgrounds? etc.), but adding debug isn't a very debatable thing, I'll only add debug in this pr.

We can do a followup afterwards with Eq.

fdehau:
You make both valid points. In addition of merging this, I will try to go through all public types and follow this [recommendation](https://rust-lang-nursery.github.io/api-guidelines/interoperability.html#c-common-traits). Regarding `Eq` for `Text`, I'm don't really expected people to compare `Text` objects since they are usually created to be given as input of a widget.

PR #156, opened 2019-05-30
Don't highlight Tabs divider after selection

nytopop:
This PR fixes the highlight behavior of `Tabs`. Currently the divider after the current selection is highlighted as well as the selection, which seems unintended to me. Feel free to close if it's not.

Before:
2019-05-29-202755_351x57_scrot

After:
2019-05-29-202733_360x53_scrot


fdehau:
Thanks, that's definitely nicer

PR #157, opened 2019-06-02
[example: user_input] Assure the cursor responds immediatel when hitt…

Byron-TW:
…ing backspace

This was discovered with the termion backend in alacritty on OSX.


fdehau:
Nice catch. Thanks for taking the time to look into this.

PR #159, opened 2019-06-05
Fix crossterm terminal size and dark gray color

scauligi:
Crossterm reports terminal size in terms of the maximum row/column _index_, as opposed to the _count_ of rows/columns.

Also updated the color mapping, as DarkGrey now exists in crossterm.


fdehau:
Thanks for catching that.

PR #160, opened 2019-06-06
Allow rendering for those who do not implement `Widget`

Byron-TW:
This minor and non-breaking change to `Terminal` enables stateful components, which I found most useful during my work on the [`disk usage analyzer`](https://github.com/Byron/dua-cli).

An example is provided, implementing a list that scrolls the way one would expect it.

During my development, I tried a lot to workaround the limitations imposed by having to use Widget, which made it hard to use stateful components. It felt 'off' and I couldn't keep going like this.

It showed that not relying on any trait opens up the whole system to any use one would want to have, including a more _React'ish_ approach which feels much more natural. And thanks to Rust, it's absolutely safe to do.

I hope you like the example, please let me know your thoughts and possibly change requests.

_PS_: For now I have created https://crates.io/crates/tui-react to keep my changes, but I realized it would be better to have some of them merged here and to find a way to contribute 'non-standard' components back to this project. I am thinking something along the lines of [`Rocket-Contrib`](https://github.com/SergioBenitez/Rocket/tree/master/contrib/lib). However, this might be off-topic here.

### DIFF
--- a/examples/stateful_component.rs
+++ b/examples/stateful_component.rs
@@ -1,0 +1,189 @@
+#[allow(dead_code)]
+mod util;
+
+use std::io;
+
+use termion::event::Key;
+use termion::input::MouseTerminal;
+use termion::raw::IntoRawMode;
+use termion::screen::AlternateScreen;
+use tui::{backend::TermionBackend, Terminal};
+
+use crate::util::event::{Event, Events};
+use tui::style::{Color, Style};
+use tui::widgets::Text;
+
+/// This list does not implement widget, but instead provides a render call taking properties.
+/// On top of that, it can keep its own state, which can conveniently be accessible to the parent application
+/// to control it. However, the component is able to write its own state and adjust it based on information
+/// it only obtains when drawing.
+///
+/// If this seems a little like React, you might be right!
+mod list {
+    use tui::{
+        buffer::Buffer,
+        layout::Rect,
+        widgets::{Block, Paragraph, Text, Widget},
+    };
+
+    /// Thanks to the state maintained in this list, it can scroll naturally.
+    /// Compare this to the `List` `Widget` in TUI, which seems to 'stick to the bottom'.
+    #[derive(Default)]
+    pub struct StatefulList {
+        /// The index at which the list last started. Used for scrolling
+        offset: usize,
+    }
+
+    #[derive(Default)]
+    pub struct StatefulListProps<'b> {
+        pub block: Option<Block<'b>>,
+        pub entry_in_view: Option<usize>,
+    }
+
+    impl StatefulList {
+        fn list_offset_for(&self, entry_in_view: Option<usize>, height: usize) -> usize {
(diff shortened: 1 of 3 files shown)


Byron-TW:
I don't know if I can do anything to start the missing CI passes. Any advice would be appreciated :) .

fdehau:
Sorry for the wait. This feedback is very interesting and I'm glad that you find this crate useful in your project.

I've tweaked your example achieving something hopefully similar to what you wanted without having to change the public api ([here](https://github.com/fdehau/tui-rs/commit/2b754c14a5326c8f59f186b92b759220bfb0ac20)). Implementing your own widget seems to get the job done but it might only work because it is a simple example. Let me know what you think.

React (or Flutter or even SwiftUI) are obviously great sources of inspiration but I've failed to find a good way to implement that kind of behavior in Rust while providing a nice api to use. Thus the concept of "consumable" widgets and the fact that you have to manage your own state seems good enough for most terminal apps. It was also a way to enforce a clear separation between the rendering  and the state management parts. But as you highlighted some UI concepts obviously involve local states and the current api might need to be improved to make them nicer to work with (e.g scrolling).

I don't have a clear opinion on a possible "contrib" project. If more people voice their sentiment on the matter I might consider it.

Byron-TW:
Thanks for having such a thorough look. The difference between the new implementation is that now, all render state is part of the widget state. This is something I want to avoid, just to keep the stateful components as stateless as possible, hence the introduction of 'render props'. The latter require a more open API of the Terminal to be able to break out of the confines of the Widget trait. In fact, whenever I tried to introduce a trait on my own I realized that in the end, these were unnecessarily limiting - in my TUI interfaces, I only work with concrete types, because I can and because it makes most sense.

I hope this provides an additional argument of opening up the Terminal interface, or makes it clearer as to why this is useful.

fdehau:
Giving this more thoughts, I think that I like the underlying idea but I'm not 100% convinced by the way to want to introduce this feature in the crate. For the sake of exploring alternatives apis for this, I've pushed a new commit in the same branch. Would you mind take a look and give me your thoughts ?

The current `Widget` trait is inherently designed for consumable widget which was one of the founding principles of this crate. So I thought of introducing a new trait `StatefulWidget` to acknowledge the differences in behavior while keeping a coherent api.

Byron-TW:
@fdehau Another trait seems nice on paper, and I have tried that in `tui-react`. Something I noticed is that the trait forces the `render` method to some sort of signature. Even with the few widgets I had, the one size didn't fit all. I was served best by just allowing to set the terminal into a state where things could render, and do so with widgets and `render` methods with signatures that were a best fit.

Personally I like the idea of supporting maybe even various traits for convenience, but also not to hide the fact that any kind of rendering is possible if the `pre` and `post` render calls are adhered to.

fdehau:
For those interested, a possible solution for stateful widgets have been introduced here #210.

PR #163, opened 2019-06-12
Fix curses backend cursor positions

defiori:
The `get_cursor` and `set_cursor` methods for the curses backend invert rows and columns, this fixes things.


fdehau:
Thanks

PR #164, opened 2019-06-14
avoid divide by zero

rustysec:
when creating or updating barchart data, if all the values are zero, we get a divide by zero. this just avoids that error and allows for all zero bars.


fdehau:
Thanks

PR #165, opened 2019-06-20
fix for canvas rendering edge cases causing overflow errors

z2oh:
In certain edge cases (like the test case provided in `tests/chart.rs`) a canvas can cause a panic. This change tests for this edge case and includes a regression test for the future.


fdehau:
Thanks :+1:

Issue #166, opened 2019-06-30
Add click events in the widgets?


ivanceras:
I look at [gocui](https://github.com/jroimartin/gocui) and seems to have support for click events on the widgets. I would want to see this implemented, is there any reason why this is not implemented. I'd try to get this to work, but I first I'd like to know if someone has tried doing and see if it's not worth adding?

fdehau:
Currently events are not handled by the crate itself. Since the current architecture is not retained, supporting mouse events can be challenging. Some possible solutions could be providing the id of the widget where a hit happen or a way to define callbacks when those events are fired. Both imply a major change in the architecture and maybe in the api, so I would consider this a possible future improvement.

ivanceras:
The fork I have https://github.com/ivanceras/itui make the widgets hold the area they are in,  instead of passing it every draw call. The event listener is also attached to the widget. Since, all the events is triggered from the application level, the way to let widgets it is the target is by checking the area if it was hit (if it is a mouseclick). All the widgets are check against the event (ie. mouseevent) to match the coordinate (x,y) if it hits the widget area and the event name/type is the same. The event listener is then emitted/called when the conditions are met. 

I have yet to release the overall code of that glue all of this. 
Meanwhile this what it looks like https://twitter.com/ivanceras/status/1150841336590786562

lgirault:
Hi, first thanks for this really cool library. Do you still consider handling click event a possible future improvement ? Perhaps even more generally having the possibily to pass an event handler ? with some defaults event handlers ? Since that widget can now be stateful, it seems the logical next step to handle behaviors like switching tabs. I'm just starting to learn rust but I'm willing to help on the subject.

fdehau:
Thanks for your interest in the library. While recent additions to the crate might ease the processing of some user inputs, we are still far from a support of click events unfortunately. I still have not made up my mind about the way to implement this. As one can see, the library is very open regarding how, when and why you decide to draw to the terminal, especially because it does not manage an event loop. It is a nice property in a lot of cases but it also falls short for others, such as click events.

Given the current architecture of the project, I would be more inclined to provide some interface to let people check whether a particular widget was clicked than having event callback on the widget themselves because the latter imply that we have access to those events.

I would also point out that I'm not familiar with all the details and innovations that have been pushed out in the field of declarative and reactive UI recently so I may not be in the best position to provide the nicest solution to this problem =). I would thus be more than happy to see informed opinions about this matter.

[...]

PR #167, opened 2019-07-04
Enable mouseevent to be passed through the stdin stream

ivanceras:
Nothing fancy, this is an initial code for supporting [mouse events on widget](https://github.com/fdehau/tui-rs/issues/166).

Eventually, I'll add a code to determine which widget is clicked based on the MouseEvent information and try to simulate the events are triggered from the widgets when the user interacts with it.

---
This change is [Reviewable](https://reviewable.io/reviews/fdehau/tui-rs/167)


Cargo.toml
```diff
@@ -1,20 +1,20 @@
 [package]
-name = "tui"
-version = "0.6.1"
-authors = ["Florian Dehau <work@fdehau.com>"]
+name = "itui"
```
longwusha: Why should we change the crate name ?

Cargo.toml
```diff
@@ -1,20 +1,20 @@
...
 license = "MIT"
 exclude = ["assets/*", ".travis.yml"]
 autoexamples = true
 edition = "2018"
 
 [badges]
-travis-ci = { repository = "fdehau/tui-rs" }
-appveyor = { repository = "fdehau/tui-rs" }
+travis-ci = { repository = "ivanceras/itui" }
+appveyor = { repository = "ivanceras/itui" }
```
longwusha: I think you have merged some wrong file.

ivanceras:
Oops, sorry. The suceedding commits are my own experiments. Don't merge it.

PR #170, opened 2019-07-10
Add percentage widths to Table widget

jeffa5:
This adds a method for using percentage widths with the table. This allows a table to span the block it is contained in rather than clinging to one side.

I added some tests for the tables too. These uncovered some bugs so I fixed them and I've written more to cover the percentage widths addition.


src/widgets/table.rs
```diff
@@ -60,6 +60,8 @@ where
     /// Width of each column (if the total width is greater than the widget width some columns may
     /// not be displayed)
     widths: &'a [u16],
+    /// Percentage width of each column, allows table to expand horizontally to fill space
+    percentage_widths: &'a [u16],
```
fdehau: Have you given thoughts about using an `Enum` that would allow `Fixed` and `Percent` width in `widths` instead ?

jeffa5: I did think about it, it would be cool to have but I wasn't sure about introducing a breaking change here.

fdehau: I already have breaking changes planned, so I'm fine with this one =)

fdehau:
Any updates on this ? I like what you did there and I would be happy to have those changes in the next release.

jeffa5:
Hi, sorry for the lack of response. I've just been on a long holiday.

I'll take a look at it in the coming days and make the change to an enum.

jeffa5:
I've made the change to an enum, let me know what you think for this.

fdehau:
Thanks the overall change looks great and I particularly liked the effort you put into properly test this. I have a last minute thought however: what about using the existing layout `Constraint` enum instead of introducing a new one ?

PR #173, opened 2019-07-20
Crossterm Performance Update

TimonPost:
I have been working on improving the performance of crosterm. The status can be followed here:https://github.com/TimonPost/crossterm/pull/175.  On windows, the performance increased with 60% and on Linux, the performance increased with +- 40% . 

This is positive for TUI, and now there are no glitches anymore in the screen. 

Before:

old

After

new

The noticeable change is, especially in the windows implementation. Windows is way slower with terminal actions. Like I had a benchmark where I did some stuff on windows it took 5 seconds to complete and on Linux it took less than one second.

For now, this should not be merged yet. I am still working on it, and once crossterm 0.10 releases I will update this PR as well.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,9 +28,9 @@ log = "0.4"
 either = "1.5"
 unicode-segmentation = "1.2"
 unicode-width = "0.1"
-termion = { version = "1.5", optional = true }
+termion = { version = "1.5" , optional = true }
 rustbox = { version = "0.11", optional = true }
-crossterm = { version = "^0.9", optional = true }
+crossterm = { version = "^0.10", optional = true }
 easycurses = { version = "0.12.2", optional = true }
 pancurses = { version = "0.16.1", optional = true, features = ["win32a"] }
 
--- a/examples/crossterm_demo.rs
+++ b/examples/crossterm_demo.rs
@@ -7,12 +7,13 @@ use std::sync::mpsc;
 use std::thread;
 use std::time::Duration;
 
-use crossterm::{input, AlternateScreen, InputEvent, KeyEvent};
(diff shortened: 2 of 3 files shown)


TimonPost:
new (2)

TimonPost:
It is ready to be merged, I updated the version of crossterm to 0.10.1

TimonPost:
How it used to be:
old_version

How it looks like with the update:

new (2)


It is ready to be released, I updated the version of crossterm

MSleepyPanda:
This is awesome! To be honest, i'm kind of appalled by the performance of powershell / cmd in general, but especially in combination with interactive terminal applications.

I'm seeing some serious memory leaks in the underlying console subsystem, see https://github.com/tokio-rs/console/issues/22 . Please ping me when this patch reaches a usable state, i will report back whether this solves it!

TimonPost:
@fdehau can this be merged?

Cargo.toml
```diff
@@ -28,9 +28,9 @@ log = "0.4"
 either = "1.5"
 unicode-segmentation = "1.2"
 unicode-width = "0.1"
-termion = { version = "1.5", optional = true }
+termion = { version = "1.0" , optional = true }
```
fdehau: why a change on the termion version ?

fdehau:
Beside my small comment, this is great work and I would be happy to merge it. Could you maybe also squash some commits to make all of them relevant ? I'm thinking that because of the breaking changes in the public api this is going to be in a 0.7 release.

TimonPost:
Awesome, I'll squash my commits, and yes it has braking changes so a new version is required.

Cargo.toml
TimonPost: Ah, good catch, I was trying something out there. But should be reverted.

TimonPost:
squashed!

PR #175, opened 2019-07-24
Make margins be vertical or horizontal

jeffa5:
This adds support for margins to be either vertical or horizontal, or
both.


fdehau:
Thanks !

PR #179, opened 2019-08-10
Typos in symbols constants

ALobsterDog:
"quarter" was misspelled as "quater" in the `symbols` module and in a few places referencing it. All I've done is replace every instance of "QUATER" in the codebase with "QUARTER." No logic changes, ran test suite just in case, all tested ok.


fdehau:
ah =) thanks !

PR #180, opened 2019-08-11
Crossterm

fdehau:
@TimonPost I've removed some compilation warnings and fixed some rendering artifacts in the crossterm demo (e.g bold everywhere at least on Linux). Some feedback would be very much appreciated if you have time.

### DIFF
--- a/src/backend/crossterm.rs
+++ b/src/backend/crossterm.rs
@@ -1,13 +1,19 @@
-use std::io;
-
-use crate::backend::Backend;
-use crate::style::{Color, Modifier};
-use crate::{buffer::Cell, layout::Rect, style};
-use crossterm::{
-    execute, queue, terminal, Clear, ClearType, Command, Crossterm, ErrorKind, Goto, Hide, Output,
-    SetAttr, SetBg, SetFg, Show,
+use std::{
+    fmt,
+    io::{self, Write},
 };
-use std::io::{stdout, Stdout, Write};
+
+use crossterm::{
+    execute, queue, terminal, Clear, ClearType, Crossterm, ErrorKind, Goto, Hide, SetAttr, SetBg,
+    SetFg, Show,
+};
+
+use crate::{
+    backend::Backend,
+    buffer::Cell,
+    layout::Rect,
+    style::{Color, Modifier, Style},
+};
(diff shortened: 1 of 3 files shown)


src/backend/crossterm.rs
```diff
@@ -81,21 +67,18 @@ where
     W: Write,
 {
     fn clear(&mut self) -> io::Result<()> {
-        queue!(self.stdout, Clear(ClearType::All));
-        self.stdout.flush();
-        Ok(())
+        queue!(self.stdout, Clear(ClearType::All))?;
+        self.stdout.flush()
```
TimonPost: Although this is good, take notice that when you flush the screen, a system call is made to flush the output buffer to the terminal screen. Queue prevents this behavior and make's you able to handle flushing your self. That might be useful in the case when you call clear 'a lot of times'. Crossterm also supports `execute` this willl execute the command for you by calling flush. 

Instead of this
```rust
        queue!(self.stdout, Clear(ClearType::All))?;
        self.stdout.flush()
```

you can do this:

```rust 
        execute!(self.stdout, Clear(ClearType::All))?;
```

src/backend/crossterm.rs
```diff
@@ -81,21 +67,18 @@ where
...
+        execute!(self.stdout, Hide)?;
+        self.stdout.flush()
     }
 
     fn show_cursor(&mut self) -> io::Result<()> {
-        execute!(self.stdout, Show);
-        self.stdout.flush();
-        Ok(())
+        execute!(self.stdout, Show)?;
+        self.stdout.flush()
```
TimonPost: execute! already does a flush, so you don't have to do this your self

src/backend/crossterm.rs
```diff
@@ -104,9 +87,8 @@ where
     }
 
     fn set_cursor(&mut self, x: u16, y: u16) -> io::Result<()> {
-        queue!(self.stdout, Goto(x, y));
-        self.stdout.flush();
-        Ok(())
+        queue!(self.stdout, Goto(x, y))?;
+        self.stdout.flush()
```
TimonPost: same here, consider using execute instead

PR #198, opened 2019-11-07
Rounded corners and Double line borders

stevensonmt:
This PR will implement a Blocky trait that defines the methods needed for drawing a block. There are three block types: Block, RoundedBlock, and DoubleBlock. The only differences are the unicode characters used to draw the borders. The RoundedBlock and DoubleBlock structs and methods will only be implemented when user applies the "rounded" and "double" features. This could avoid complicating things if simplicity is the goal.

### DIFF
--- a/examples/block.rs
+++ b/examples/block.rs
@@ -9,7 +9,7 @@ use termion::screen::AlternateScreen;
 use tui::backend::TermionBackend;
 use tui::layout::{Constraint, Direction, Layout};
 use tui::style::{Color, Modifier, Style};
-use tui::widgets::{Block, Borders, Widget};
+use tui::widgets::{Block, BorderType, Borders, Widget};
 use tui::Terminal;
 
 use crate::util::event::{Event, Events};
@@ -35,7 +35,7 @@ fn main() -> Result<(), failure::Error> {
             Block::default()
                 .borders(Borders::ALL)
                 .title("Main block with round corners")
-                .rounded()
+                .set_border_type(BorderType::Rounded)
                 .render(&mut f, size);
             let chunks = Layout::default()
                 .direction(Direction::Vertical)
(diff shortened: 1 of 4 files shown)


stevensonmt:
Running the termion example with `cargo run --features rounded --features double --example termion_demo should give you rounded corners for Graphs and List with double borders for Tabs and Bar Chart:
Screenshot_2019-11-06_20-15-29

fdehau:
I appreciate the effort but the implementation seems quite complex for such feature. Have you given any thoughts on having an enum for the corners and for whether the border is doubled or not ? It could look like this:
```rust
enum Corners {
  Straight,
  Rounded,
}
enum Lines {
   Simple,
   Double
}
```
to be used like this:
```rust
Block::default().corners(Corners::Rounded).lines(Lines::Double);
```
This removes the need for feature flags and only the `Block` widget would have to change.

stevensonmt:
I'm happy to redo the feature in that style. It seemed less challenging to avoiding breaking changes to do it this way. Changing the existing Block widget might mean that it is not backwards compatible.

stevensonmt:
yeah, actually I still don't know how to implement it in a non-breaking way other than the approach I took. I can add a field to the Block struct that specifies the border line style, but that is potentially breaking for anyone whose code does not contain that field in their Block instances.
I can add a parameter to the `draw` function but again, if code does not contain that parameter it will be a breaking change.

stevensonmt:
Sorry about the one sort of nonsensical commit. Had a git rebase issue I wasn't sure how to fix. At any rate, the block example now uses both double lined borders and rounded corners.

fdehau:
I don't think you have to worry about any breaking changes. If I take my previous proposition, `Block::default` would take care of initializing the new properties in a backward compatible way and the new methods `corners` and `lines` could be used by developers to modify those properties. The only big change would be internal in the `Block::draw` method.

stevensonmt:
Okay, I'll try it with adding a `border_type` field that will take an enum that is `Plain | Rounded | Double` and set it to `Plain` in the default impl. The `draw` method will then match on that enum to get the appropriate symbols.

PR #205, opened 2019-11-30
Added a verticle 'cross' to the symbols as it was missing

kyruzic:
An intersecting column and row where the row is not at the top or bottom intersect as a cross, the symbols::line module has every other intersection of the line so it makes sense that this should be added as well. 

Attached is my use case for this. Notice the cross symbol highlighted in red. 
2019-11-30-145358_1256x680_scrot


fdehau:
Thanks ! As I suspect that you have implemented your own widget, can I encourage you to open a followup PR to make those borders available in the`Table` widget ? They look nice in your screenshot and I'm pretty sure somebody has already ask for them.

kyruzic:
Hopefully I will have some time in the near future to get that done. But I'm quite busy currently.

PR #212, opened 2020-01-03
docs(readme): add bandwhich to "apps using tui"

imsnif:
Thanks very much for this library! :)


fdehau:
Glad to see this. That's a very cool project.

Issue #214, opened 2020-01-09
user_input example quits on "q"


nphyx:
**Describe the bug**
Trivial issue - the user_input example uses "q" as the exit key, so you can't type anything with the letter "q".

Maybe make the exit character [esc] or [Ctrl+c]. Wouldn't hurt to do the same in other examples.

**To Reproduce**
1) `cargo run --example user_input`
2) type a word with q, like "quintessence"

**Expected behavior**
The input is accepted.

**Desktop**
 - OS: Linux
 - Terminal Emulator: xfce terminal
 - Font: Monofur (+nerd font mods)
 - Crate version: 0.8.0
 - Backend: termion

**Additional context**
I'd be happy to make a PR with this change (and to the other examples if you'd like).

fdehau:
Thanks for the report. It was fixed in e6ce0ab9a7f4f71c42f53fc7c34eda0d78eb46aa using input modes.

qiujiangkun:
This problem still exists in `tui v0.14.0`

fdehau:
@qiujiangkun I've just tested and I've been able to input message with the letter "q". Could you give more details on what is not working as expected ?

qiujiangkun:
```rust
impl Default for Config {
    fn default() -> Config {
        Config {
            exit_key: Key::Char('q'),
            tick_rate: Duration::from_millis(250),
        }
    }
}
```
Default config assumes 'q' as exit. I didn't notice this 4 days ago.

fdehau:
I still don't see how this is related to the original issue though, there is no problem in the `user_input` example, right ?

qiujiangkun:
Yes, the examples are fine

PR #216, opened 2020-01-10
Fixed user_input example exit key

dylanwishner:
Fixes issue #214 - Changed the default key on user_input example to 'Esc' rather than 'q' to prevent the program from quitting when users enter a word that contains 'q'


fdehau:
Closing in favor of e6ce0ab9a7f4f71c42f53fc7c34eda0d78eb46aa because I rather show an example of input modes and have the same exit key for all examples.

PR #217, opened 2020-01-11
Add linechart support

### LINKED ISSUE #73: Feature request: Linechart
I think it would be nice if we had an option for the chart widget to draw lines between each marker. I saw you were able to have a linechart in `examples/demo.rs` but only when feeding in continuous data. Maybe add an option to the chart widget or create a new linechart struct to enable drawing a line between consecutive points. I could probably hack a pr together but I wasn't sure what the best way to implement it would be.

cjbassi:
Closes #73

This commit only adds support for linecharts for the braille marker.

### DIFF
--- a/src/widgets/chart.rs
+++ b/src/widgets/chart.rs
@@ -6,7 +6,7 @@ use crate::buffer::Buffer;
 use crate::layout::Rect;
 use crate::style::Style;
 use crate::symbols;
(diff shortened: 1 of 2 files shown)


fdehau:
There is a merge conflict. Otherwise this looks good to me =).

cjbassi:
Fixed!

PR #220, opened 2020-01-14
Add show_baseline field to sparkline

cjbassi:


fdehau:
Could you describe a bit what you are trying to achieve here ? From my understanding, your changes make sure that there will always be at least one eight of a block if `show_baseline` is set. But for me that makes different values have the same representation. Wouldn't it be more correct and simpler to apply the same offset to all the computed data ?

```rust
let offset = if self.show_baseline { 1 } else { 0 };
let mut data = self
            .data
            .iter()
            .take(max_index)
            .map(|e| {
                if max != 0 {
                    e * u64::from(spark_area.height) * 8 / max + offset
                } else {
                    0
                }
            })
            .collect::<Vec<u64>>();
```

cjbassi:
Yeah, so line 91 draws the baseline across the entire width of the sparkline, and line 120 makes sure the baseline doesn't get overwritten when drawing the data, although I think you're right and line 120 needs to be fixed.

Issue #196, opened 2019-11-05
`SelectableTable` - Feature Request


sputnick1124:
Recently in `spotify-tui`, we had a need to have a table which would automatically scroll to allow the highlighted item in the table always visible, similar to how `SelectableList` works. 

[This was implemented](https://github.com/Rigellute/spotify-tui/pull/135) for the use case within `spotify-tui`, but perhaps would be widely applicable to other users of `tui-rs`.

I'd love to get feedback from other users if this would be broadly useful. I could even take a shot at an implementation if so. Thanks!

[![asciicast](https://asciinema.org/a/FdKDHNjloi0CgFKXo7etLvYjE.svg)](https://asciinema.org/a/FdKDHNjloi0CgFKXo7etLvYjE)

aryakaul:
Just wanted to drop a note saying that at least I would find the Widget helpful.

I'm currently trying to adapt your implementation to address the use case described in #200 -- really cool work!

orhun:
It's a great idea, must be implemented. I'm using your snippet that handles the selection with an offset and `skip` function. Working well!

fdehau:
I would be more than happy to have such feature in the library. I would also suggest to merge this functionality directly in the `Table` widget if possible (as it is my intent to merge `List` and `SelectableList`). PRs are welcome.

sputnick1124:
Sounds good! I'll start putting some code together in the coming days when I get some down time over the holidays.

sputnick1124:
Sorry this took me so long to get around to. I got seriously distracted by Advent of Code :blush:

fdehau:
It has been implemented in #242 using the initial work of @sputnick1124.

PR #222, opened 2020-01-17
Selectable table

sputnick1124:
This adds the option to have a selected item in a `Table` which can be highlighted with a style and symbol (a parallel feature with the `SelectableList` type). The constructor was not changed to avoid breaking existing code. Selectability can be configured using the builder pattern as is common in the rest of the code base.

[asciicast](https://asciinema.org/a/qf6RCoFQL0pn934C2YyySTx5S)

Closes https://github.com/fdehau/tui-rs/issues/196


laumann:
@sputnick1124 Thank you! I've wanted this feature for some time, but never dug into the implementation myself. I imagine though that you might have to redo the implementation to make it a `StatefulWidget`. (`SelectableList` has been removed.)

sputnick1124:
Glad it'll be useful! I'll take a look at it when I get some time later this month to see what needs to be changed. To be clear, I didn't use `SelectableList` in my implementation, but baked selectability into the current implementation of the `Table` widget. At and rate, this was some time ago now and I assume the PR will need some updating at least.

laumann:
True, I realized that after leaving my comment :-) But I think there's a little more work to be done anyhow. I'm updating my codebase to run off of `master` right now, because I want to see how the `StatefulWidget` trait stuff works out.

fdehau:
@sputnick1124 I built upon your work here to implement the feature in #242. Thanks again for your work and sorry for the late response/review.

PR #226, opened 2020-01-20
Mark Style::* functions `const`

### LINKED ISSUE #223: Make Style::* functions `const`
**Is your feature request related to a problem? Please describe.**
I would like to declare `const`ants for a few styles, which could then be used everywhere easily.
```rust
    const SOME_STYLE: Style = Style::new()
        .fg(Color::Red)
        .bg(Color::Blue)
        .modifier(Modifier::DIM);
```
Currently I have to utilize the lazy_static crate for that.

**Describe the solution you'd like**
Adding `const` to:
- `Style::fg`
- `Style::bg`
- `Style::modifier`
and adding a `pub const fn new() -> Self` which does the same as `Style::default` would allow that.

**Describe alternatives you've considered**
[lazy_static](https://crates.io/crates/lazy_static), which is unnecessary overhead.

I already tested my solution and could assemble a PR..

MalteT:
Fix #223.

I tested the changes with the following snippet, just to verify that the changes actually enabled the use
of `const FOO: Style...`. Since the test is quite useless I removed them from the PR.

```rust
#[cfg(test)]
mod tests {
    use super::*;

    const RED: Color = Color::Red;
    const RED_STYLE: Style = Style::new().fg(RED);
    const RED_DIMMED: Style = Style::new()
        .fg(Color::Red)
        .bg(Color::Blue)
        .modifier(Modifier::DIM);

    #[test]
    fn test_const() {
        assert_eq!(RED, Color::Red);
        assert_eq!(
            RED_STYLE,
            Style {
                fg: Color::Red,
                ..Style::default()
            }
        );
        assert_eq!(
            RED_DIMMED,
            Style {
                fg: Color::Red,
                bg: Color::Blue,
                modifier: Modifier::DIM,
            }
        );
    }
}
```

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -114,30 +114,33 @@ pub struct Style {
 
 impl Default for Style {
     fn default() -> Style {
+        Style::new()
+    }
+}
+
+impl Style {
+    pub const fn new() -> Self {
         Style {
             fg: Color::Reset,
             bg: Color::Reset,
             modifier: Modifier::empty(),
         }
     }
-}
-
-impl Style {
     pub fn reset(&mut self) {
         self.fg = Color::Reset;
         self.bg = Color::Reset;
         self.modifier = Modifier::empty();
     }
 
-    pub fn fg(mut self, color: Color) -> Style {
+    pub const fn fg(mut self, color: Color) -> Style {
         self.fg = color;
         self
     }
-    pub fn bg(mut self, color: Color) -> Style {
+    pub const fn bg(mut self, color: Color) -> Style {
         self.bg = color;
         self
     }
-    pub fn modifier(mut self, modifier: Modifier) -> Style {
+    pub const fn modifier(mut self, modifier: Modifier) -> Style {
         self.modifier = modifier;
         self
     }
--- a/tests/paragraph.rs
+++ b/tests/paragraph.rs
@@ -4,8 +4,7 @@ use tui::layout::Alignment;
 use tui::widgets::{Block, Borders, Paragraph, Text, Widget};
 use tui::Terminal;
 
-const SAMPLE_STRING: &str =
-    "The library is based on the principle of immediate rendering with \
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "tui"
-version = "0.8.0"
+version = "0.8.1"
```
cjbassi: The version should probably not be bumped here.

MalteT: Can I somehow drop that commit from the PR or do I open another one?

cjbassi:
Would it make sense to change all of the builder methods across the entire project to const and not just for the Style?

Cargo.toml
cjbassi: I would just drop that commit from the PR. If you need help with that then let me know :)

MalteT: I did a force push to my fork now, is there a better alternative? And this is probably off-topic, but could you explain to me, why the version bump is not necessary?

MalteT:
I could mark alot more functions as `const`. These were just the ones I found to be most annoying. Is there any downside to this, @cjbassi?

cjbassi:
Yeah that makes sense, I was just wondering. It should be fine to just do these ones.

Cargo.toml
cjbassi: Yeah, that's the way I would do it.

There's no reason to bump the version number since that's more of the maintainers job and it's easy for them to do. Plus, you wouldn't want to create conflicts with other pull requests that do the same.

MalteT: Well, that makes sense. Thank You!

fdehau:
While I see how having `const` Style methods can be useful for the reasons mentioned in the issue, I'm a bit skeptical on moving all builder methods to const. The widgets are meant as consumable objects so I can't think of an example where this would be useful. Feel free to open a distinct issue though =).

PR #231, opened 2020-02-17
Add "thick" lines and line::Set struct

lgirault:
Add a new style of line and use a struct to avoir duplication of
matching

### DIFF
--- a/src/symbols.rs
+++ b/src/symbols.rs
@@ -23,40 +23,115 @@ pub mod bar {
 pub mod line {
     pub const VERTICAL: &str = "│";
     pub const DOUBLE_VERTICAL: &str = "║";
+    pub const THICK_VERTICAL: &str = "┃";
 
     pub const HORIZONTAL: &str = "─";
     pub const DOUBLE_HORIZONTAL: &str = "═";
+    pub const THICK_HORIZONTAL: &str = "━";
 
     pub const TOP_RIGHT: &str = "┐";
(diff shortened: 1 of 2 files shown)


fdehau:
I really like your approach though it got me thinking that we could directly specify the values in the structs you introduced and take advantage of the namespace associated with the modules.
What do you think of ?

```rust
pub mod lines {
     pub struct Set {
          vertical: &'static str,
          horizontal: &'static str,
     }
     pub const NORMAL: Set {
          vertical: "|",
          horizontal: "─",
          ..
     }
     pub const ROUNDED: Set {
        ..
     }
     pub const DOUBLE: Set {
       ..
     }
     pub const THICK: Set {
       ..
     }
```
And is it me or "thick" is a more correct adjective for the kind of line you want to add ?

lgirault:
I didn't want to change too much the existing code and since some structs share most of the characters (NORMAL & ROUNDED) I thought it made sens to be able to access symbols from both way. But I can see how it feels redundant and I'd be happy to modify the PR to directly specify the values in the structs.
You're completely right that thick is a more correct adjective. English is not my primary langage, hence the error.

lgirault:
I almost replaced the symbol consts with the Set structs but if I'm not mistaken the field of a struct, even a const, cannot be used as a pattern in a match. So I left both declarations.

fdehau:
That makes perfect sense. I merge as it is then.

PR #233, opened 2020-02-24
docs: update to build more backends

cardoe:
Enable more backends to be built into the default docs than just
termion. These were the only backends that built in a representative
environment of docs.rs. Options come from https://docs.rs/about#metadata

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,10 @@ rand = "0.7"
 failure = "0.1"
 structopt = "0.3"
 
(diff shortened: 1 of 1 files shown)


fdehau:
Thanks for pointing this out, I was not aware of this feature. However, I suspect that it will not work for the curses features as the corresponding backend requires `libncurses5-dev` to be installed. Do you know if it is possible to test that behavior beforehand (like a tool that replicate what docs.rs does behind the scene at the build step) ?

PR #234, opened 2020-02-25
Clean redundant generics params in Table

vchekan:
It is possible to use associated types the same way as generics parameters. In fact associated types are nothing more than better organized generics params. For example, there is no need to introduce another param to constraint iterator item to be Display, you can just say `where I::Iterator, I::Item: Display`. This allows to drop type params for Table from 5 to 2.


fdehau:
That's indeed a lot nicer. Thanks!

PR #236, opened 2020-03-02
Add zenith to apps using tui

bvaisvil:
Using `tui-rs` as been great and I'd like to add my app zenith to the `Apps using Tui` section.


fdehau:
Always glad to see what people are doing with this crate =)

Issue #211, opened 2019-12-30
 Add Popup box


betta-cyber:
It would be nice to have a popup box widget or a dialog widget to show some message for user.

![alt image](https://camo.githubusercontent.com/1f2e44ca0074c3c5b06880cab3b2062595693d1c/68747470733a2f2f692e696d6775722e636f6d2f33596c656f7a632e706e67)

It could be like this. cursive have this widget. so i choose it for a example.

fdehau:
closing because @extrawurst has shown in #251 how to implement this.

dataO1:
I think it would still be great to have this in the standard toolbox of Widgets.
I needed to google and find this issue to know, that someone has already implemented this, I think having it on the github readme would be great!

PR #244, opened 2020-03-14
feat: Add clean option to block

heyrict:
Related to: #211 

I find the need to make a pop-up dialog in my program, but drawing widgets over drawn buffer will not clean all the underlying cells. Thus I add a `reset_area` function to `buffer.rs` to reset cells in a certain `Rect`. This will make it possible to implement a dialog. 

I also adds a `clean` option to `Block` to clean the cells first before rendering. Other widgets are untouched.

Screenshot from 2020-03-14 13-39-03

Here is a minimal-example to create a dialog:
```rust
content = frame.size();

// The main view
frame.render_widget(
    MainViewWidget::new(self.app),
    content,
);

// The dialogue
if self.app.show_save_model {
    let modal_messages: [Text, _] = ...;
    let modal_buttons: [Text, _] = ...;
    let messages_blk: Block = ...;
    let buttons_blk: Block = ...;
    let (margin_x, margin_y) = calculate_margins(...);

    let layout = Layout::default()
        .constraints([Constraint::Length(6), Constraint::Length(2)].as_ref())
        .vertical_margin(margin_y)
        .horizontal_margin(margin_x)
        .split(content);

    frame.render_widget(
        Paragraph::new(modal_messages.iter()).block(messages_blk.clean(true)),
        layout[0],
    )
    frame.render_widget(
        Buttons::new(buttons_group.iter()).block(buttons_blk.clean(true)),
        layout[1],
    )
}
```

### DIFF
--- a/src/widgets/block.rs
+++ b/src/widgets/block.rs
@@ -42,6 +42,8 @@ pub struct Block<'a> {
     border_type: BorderType,
     /// Widget style
     style: Style,
+    /// Whether to reset the background
+    clean: bool,
 }
 
 impl<'a> Default for Block<'a> {
@@ -53,6 +55,7 @@ impl<'a> Default for Block<'a> {
             border_style: Default::default(),
             border_type: BorderType::Plain,
             style: Default::default(),
+            clean: false,
         }
     }
 }
@@ -88,6 +91,11 @@ impl<'a> Block<'a> {
         self
     }
 
+    pub fn clean(mut self, clean: bool) -> Block<'a> {
+        self.clean = clean;
+        self
+    }
+
     /// Compute the inner area of a block based on its border visibility rules.
     pub fn inner(&self, area: Rect) -> Rect {
         if area.width < 2 || area.height < 2 {
@@ -119,6 +127,10 @@ impl<'a> Widget for Block<'a> {
         }
 
         buf.set_background(area, self.style.bg);
+
+        if self.clean {
+            buf.reset_area(area);
+        }
 
         // Sides
         if self.borders.intersects(Borders::LEFT) {
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -367,6 +367,15 @@ impl Buffer {
         }
     }
 
+    /// Reset cells in given area in the buffer
+    pub fn reset_area(&mut self, area: Rect) {
+        for y in area.top()..area.bottom() {
+            for x in area.left()..area.right() {
+                self.get_mut(x, y).reset();
+            }
+        }
+    }
+
     /// Merge an other buffer into this one
     pub fn merge(&mut self, other: &Buffer) {
         let area = self.area.union(other.area);


src/widgets/block.rs
```diff
@@ -88,6 +91,11 @@ impl<'a> Block<'a> {
         self
     }
 
+    pub fn clean(mut self, clean: bool) -> Block<'a> {
+        self.clean = clean;
+        self
+    }
+
```
heyrict:
```suggestion
    pub fn clean(mut sel) -> Block<'a> {
        self.clean = true;
        self
    }
```
Maybe this is better?

Extrawurst:
I just stumbled on the same challenge. I for now took a different route:

```
pub struct Clear<T: Widget>(T);

impl<T: Widget> Clear<T> {
    pub fn new(w: T) -> Self {
        Self(w)
    }
}

impl<T: Widget> Widget for Clear<T> {
    fn draw(&mut self, area: Rect, buf: &mut Buffer) {
        if area.width < 2 || area.height < 2 {
            return;
        }

        for x in area.left()..area.right() {
            for y in area.top()..area.bottom() {
                buf.get_mut(x, y).reset();
            }
        }

        self.0.draw(area, buf);
    }
}
```

this `Clear` Widget can wrap any other widget and clears out the buffer before rendering the wrapped widget.

see example use:
```
Clear::new(
            Paragraph::new(txt.iter())
                .block(Block::default().title("Popup").borders(Borders::ALL))
                .alignment(Alignment::Center),
        )
        .render(f, Rect::new(20, 30, 100, 10));
```


But yea I think it shows how necessary a *official* solution is!

fdehau:
I think that @Extrawurst solution might be preferable because it works with all widgets. To make it behave as any other widget we might want to have it render to the same area instead of wrapping a widget instance.
```rust
let area = Rect::new(20, 30, 100, 10);
let popup = Paragraph::new(txt.iter())
   .block(Block::default().title("Popup").borders(Borders::ALL))
   .alignment(Alignment::Center);
f.render_widget(Clear, area);
f.render_widget(popup, area);
```
What do you think ?

heyrict:
So do I! @Extrawurst can you make a PR ?

Extrawurst:
Sure I can do that!

heyrict:
Closed in favor of #251

manishsingh10895:
> I think that @extrawurst solution might be preferable because it works with all widgets. To make it behave as any other widget we might want to have it render to the same area instead of wrapping a widget instance.
> 
> ```rust
> let area = Rect::new(20, 30, 100, 10);
> let popup = Paragraph::new(txt.iter())
>    .block(Block::default().title("Popup").borders(Borders::ALL))
>    .alignment(Alignment::Center);
> f.render_widget(Clear, area);
> f.render_widget(popup, area);
> ```
> 
> What do you think ?

This works pretty good, but can you help me with, rendering it on the top, like `z-index` in css ?

PR #252, opened 2020-03-25
fix: Handle resize event in crossterm demo

### LINKED ISSUE #250: crossterm (re)size broken
**Describe the bug**
In order to prepare the pull request for the Clear-feature (https://github.com/fdehau/tui-rs/pull/244) I pulled in the master and realized that crossterm terminal sizing seems broken: (see screenshot)

The terminal seems to have a default size and not using up the actual terminal size.
Also the resize events are not changing anything.

**To Reproduce**
In my case (macOS) I only need to run the Crossterm-demo:
```
cargo run --example crossterm_demo --no-default-features --features="crossterm" --release -- -tick-rate 200
```
**Expected behavior**

**Screenshots**
image

**Desktop (please complete the following information):**
 - OS: macos
 - Terminal Emulator: iTerm2
 - Crate version: master
 - Backend: crossterm

**Additional context**
This was working as intended in the previous release

heyrict:
This PR pipes `crossterm::Event::Resize` to the ui, forcing the ui to re-render after resize.

Fixes #250.

### DIFF
--- a/examples/crossterm_demo.rs
+++ b/examples/crossterm_demo.rs
@@ -21,6 +21,7 @@ use tui::{backend::CrosstermBackend, Terminal};
 
 enum Event<I> {
     Input(I),
+    Resize,
     Tick,
 }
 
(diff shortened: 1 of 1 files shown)


extrawurst:
Hi @heyrict thanks for following up on #250 but I checked this out on my machine and it does not fix the issue. it turns out that when I checkout your fork it works because for some reason it compiles against Crossterm `0.10` while my local repo uses `0.16` - please try your change against recent Crossterm and I think u will see the issue aswell.

The main issue that this shows is: the `Cargo.lock` file is on `.gitignore` which allows everyone to build against different sets of versions. Is there any reason for this @fdehau ?

fdehau:
@extrawurst The reason for the missing Cargo.lock: https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

PR #255, opened 2020-03-28
AnsiBuffer for preformatted text

lgirault:
Hi, This si a preliminar support for ansi escape code using termit-ansi for parsing.
I want to know if you are interrested by such a feature. 

If so I'll need some feedback: I'm still a newbie at rust and I'm not very happy with all the heap allocation going in the AnsiHandler, but I don't know any better since lifetime annotation are not used in the trait ...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ either = "1.5"
 unicode-segmentation = "1.2"
 unicode-width = "0.1"
 termion = { version = "1.5" , optional = true }
+termit-ansi = "0.2.1"
 rustbox = { version = "0.11", optional = true }
 crossterm = { git = "https://github.com/crossterm-rs/crossterm", optional = true }
 easycurses = { version = "0.12.2", optional = true }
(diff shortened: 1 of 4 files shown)


fdehau:
I think I understand what you are trying to achieve with this but I'm not sure whether this functionality should live in this crate. Since this crate is abstracting away the ansi escape sequences and has settled on the `Text` and `Style`, transforming the input to those structs is not really the responsibility of this crate but of your application. If it was a common enough issue, this could also be implemented in a third-party widget.

lgirault:
So my uses case is that I'm writing a mud-client and I can receive color-formatted text. 
Handling this style is not the concern of the mud domain but rather of the ui. At that moment, the feature looking generic enough, it seemed natural to make a PR. However now that you point it, I see how it is paradoxal that on one hand the whole library abstract the ansi code while on the other hand there is this bizarre feature that propose to directly interpret the ansi character. 
I think the idea of the third-party widget is what makes more sens.

fdehau:
I'm closing since we kind of agree on the whether it should be part of this crate. I would happily link to a third-party widget or to some examples showing how it could be done if you managed to put something together.

PR #270, opened 2020-04-15
fix(widgets/list): fix line length calculation for selectable lists

dotdash:
The code that outputs the list elements uses the full inner width of its
block, without taking the width of the highlight symbol into
consideration. This allows the elements to overflow the box and draw
over the block's border. To fix that, we need to reduce the target width
for the list elements.


fdehau:
Thanks !

PR #274, opened 2020-04-18
docs: Fix missing code block fence for hidden_legend_constraints

ClementTsang:
Two things:

1.  Minor thing that I noticed when using this function.  Seems to look fine in documentation, but looked weird on stuff like VS Code.

image

2.  When using `GraphType::Line`, if the dataset is empty, you can end up with a subtraction underflow when you do the length - 1.

### DIFF
--- a/src/widgets/chart.rs
+++ b/src/widgets/chart.rs
@@ -304,6 +304,7 @@ where
     /// // or if its height is greater than 25% of the total widget height.
     /// let _chart: Chart<String, String> = Chart::default()
     ///     .hidden_legend_constraints(constraints);
+    /// ```
     pub fn hidden_legend_constraints(
         mut self,
         constraints: (Constraint, Constraint),
--- a/src/widgets/chart.rs
+++ b/src/widgets/chart.rs
@@ -500,14 +500,16 @@ where
                         color: dataset.style.fg,
                     });
                     if let GraphType::Line = dataset.graph_type {
-                        for i in 0..dataset.data.len() - 1 {
-                            ctx.draw(&Line {
-                                x1: dataset.data[i].0,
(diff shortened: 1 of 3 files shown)


ClementTsang:
EDIT: This was originally just a documentation fix but I found the `GraphType::Line` bug soon after... I guess kill two birds with one PR.

fdehau:
I think that your fix is the object of an other dedicated PR (#283). So I would happily take the documentation fix in this PR and resolve the overflow bug in the other one.

ClementTsang:
Ah, my bad.  Forgot that I pushed that change temporarily to use in my Cargo.  I'll revert the third commit.

The second commit is more of a fix for if `dataset.data` is empty - `for i in 0..dataset.data.len() - 1 {...}` occasionally seems to fail otherwise due to underflow.

src/widgets/chart.rs
```diff
@@ -499,14 +500,16 @@ where
...
-                        for i in 0..dataset.data.len() - 1 {
-                            ctx.draw(&Line {
-                                x1: dataset.data[i].0,
-                                y1: dataset.data[i].1,
-                                x2: dataset.data[i + 1].0,
-                                y2: dataset.data[i + 1].1,
-                                color: dataset.style.fg,
-                            })
+                        if !dataset.data.is_empty() {
+                            for i in 0..dataset.data.len() - 1 {
```
fdehau, line 504: I think that we could use `slice::windows` instead which handles the empty case nicely.

fdehau:
Thanks. No you can left the other bug fix here that's fine. I've left a comment and it would be nice if you could add a small test covering this case as well.

fdehau:
Thanks !

PR #276, opened 2020-04-22
Add desed to list of apps using tui?

SoptikHa2:
Desed is new application that is using tui-rs. It's debugger for sed with all the things like stepping, setting breakpoints and examining runtime state.

It passed 500 github stars, so I think it's big and popular enough to be added to the list in readme.


fdehau:
This is a nice project ! I might start using it ;).

PR #277, opened 2020-04-22
add documentation field

chrunchyjesus:
With this, the `documentation` link should also appear in the search result page of `crates.io` when searching for `tui`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ authors = ["Florian Dehau <work@fdehau.com>"]
 description = """
 A library to build rich terminal user interfaces or dashboards
 """
-documentation = "https://docs.rs/tui/latest/tui"
+documentation = "https://docs.rs/tui/0.9.1/tui/"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -5,6 +5,7 @@ authors = ["Florian Dehau <work@fdehau.com>"]
 description = """
 A library to build rich terminal user interfaces or dashboards
 """
+documentation = "https://docs.rs/tui/latest/tui"
```
fdehau, line 8: Shouldn't the `latest` replaced by the current version of the crate ?

chrunchyjesus: Well, yes. But then it needs to be updated every time there is a new release.

fdehau: I'm aware of that but since this is effectively packaged with the release, I assume that it would be better that this link was in sync.

chrunchyjesus: Fixed.

fdehau: Thanks

Issue #280, opened 2020-04-27
Using line generation can cause panics due to subtraction with overflow


ClementTsang:
**Describe the bug**

Was eager to use the new line feature for graphs to avoid having to manually do the linear interpolation.  Unfortunately, using it has caused this following panic to show up after running for a bit:

```
thread '<unnamed>' panicked at 'attempt to subtract with overflow', /home/blahaj/.cargo/registry/src/github.com-1ecc6299db9ec823/tui-0.9.1/src/widgets/canvas/line.rs:82:30
   0: btm::panic_hook
             at src/main.rs:532
   1: btm::main::{{closure}}
             at src/main.rs:155
   2: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:474
   3: rust_begin_unwind
             at src/libstd/panicking.rs:378
   4: core::panicking::panic_fmt
             at src/libcore/panicking.rs:85
   5: core::panicking::panic
             at src/libcore/panicking.rs:52
   6: tui::widgets::canvas::line::draw_line_high
             at /home/blahaj/.cargo/registry/src/github.com-1ecc6299db9ec823/tui-0.9.1/src/widgets/canvas/line.rs:82
   7: <tui::widgets::canvas::line::Line as tui::widgets::canvas::Shape>::draw
             at /home/blahaj/.cargo/registry/src/github.com-1ecc6299db9ec823/tui-0.9.1/src/widgets/canvas/line.rs:52
   8: tui::widgets::canvas::Context::draw
             at /home/blahaj/.cargo/registry/src/github.com-1ecc6299db9ec823/tui-0.9.1/src/widgets/canvas/mod.rs:279
   9: <tui::widgets::chart::Chart<LX,LY> as tui::widgets::Widget>::render::{{closure}}
             at /home/blahaj/.cargo/registry/src/github.com-1ecc6299db9ec823/tui-0.9.1/src/widgets/chart.rs:503
  10: <tui::widgets::canvas::Canvas<F> as tui::widgets::Widget>::render
             at /home/blahaj/.cargo/registry/src/github.com-1ecc6299db9ec823/tui-0.9.1/src/widgets/canvas/mod.rs:443
  11: <tui::widgets::chart::Chart<LX,LY> as tui::widgets::Widget>::render
             at /home/blahaj/.cargo/registry/src/github.com-1ecc6299db9ec823/tui-0.9.1/src/widgets/chart.rs:491
  12: tui::terminal::Frame<B>::render_widget
[...]

PR #283, opened 2020-05-06
Add bounds check when drawing line high/low

DarrienG:
Fix for underflow bug when using line generation: https://github.com/fdehau/tui-rs/issues/280

### DIFF
--- a/src/widgets/canvas/line.rs
+++ b/src/widgets/canvas/line.rs
@@ -64,7 +64,11 @@ fn draw_line_low(painter: &mut Painter, x1: usize, y1: usize, x2: usize, y2: usi
     for x in x1..=x2 {
         painter.paint(x, y, color);
         if d > 0 {
-            y = if y1 > y2 { y - 1 } else { y + 1 };
+            y = if y1 > y2 {
+                y.saturating_sub(1)
+            } else {
+                y.saturating_add(1)
+            };
             d -= 2 * dx;
         }
         d += 2 * dy;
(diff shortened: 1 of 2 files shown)


src/widgets/canvas/line.rs
```diff
@@ -64,7 +64,7 @@ fn draw_line_low(painter: &mut Painter, x1: usize, y1: usize, x2: usize, y2: usi
     for x in x1..=x2 {
         painter.paint(x, y, color);
         if d > 0 {
-            y = if y1 > y2 { y - 1 } else { y + 1 };
+            y = if y1 > y2 && y1 > 0 { y - 1 } else { y + 1 };
```
extrawurst, line 67: what about the *theoretical* overflow case of `y+1` ? would it be clearer communication-wise to use `saturated_add`/`saturated_sub` here?

ClementTsang: I personally also think just using `saturated_sub/saturated_add` will do the trick.

DarrienG: TIL these exist. I'll update after I finish up after work. Thanks for the suggestions!

DarrienG: Saturated!

fdehau:
Thanks for reporting the issue. The fix makes sense but could you also add a test to cover this edge case ?

DarrienG:
Totally! I'm honestly not certain how the code path is hit, but I added a test that recreates the conditions I got the crash. If there's a better way to trigger the error, I can revise :)

PR #296, opened 2020-05-21
Borrow layout for splitting instead of moving it

knutwalker:
This allows for layouts to be saved and reused and does not require an additional clone


fdehau:
That seems reasonable =). But this will only available in the next release as it is a minor but still breaking change. Thanks.

PR #310, opened 2020-05-28
Support for headerless tables without header gap

capgelka:
I  found it useful to create a table without a header when one needs just formatted columns. This can be done by just filling the header field with empty strings, however we still have a header gap in this case.

This PR offers a solution to the problem. Maybe it will be better just to remove `1` here: https://github.com/fdehau/tui-rs/blob/master/src/widgets/table.rs#L290 or change `header` method to use `Option<T>` directly, unfortunately, this breaks the current behavior, and the solution in PR seems doesn't break anything.

### DIFF
--- a/src/widgets/table.rs
+++ b/src/widgets/table.rs
@@ -86,7 +86,7 @@ pub struct Table<'a, H, R> {
     /// Base style for the widget
     style: Style,
     /// Header row for all columns
-    header: H,
+    header: Option<H>,
     /// Style for the header
     header_style: Style,
     /// Width constraints for each column
@@ -112,7 +112,7 @@ where
         Table {
             block: None,
             style: Style::default(),
-            header: H::default(),
+            header: Some(H::default()),
             header_style: Style::default(),
             widths: &[],
             column_spacing: 1,
@@ -134,7 +134,7 @@ where
         Table {
             block: None,
             style: Style::default(),
-            header,
+            header: Some(header),
             header_style: Style::default(),
             widths: &[],
             column_spacing: 1,
@@ -153,7 +153,12 @@ where
     where
         II: IntoIterator<Item = H::Item, IntoIter = H>,
     {
-        self.header = header.into_iter();
+        self.header = Some(header.into_iter());
+        self
+    }
+
+    pub fn headerless(mut self) -> Table<'a, H, R> {
+        self.header = None;
         self
     }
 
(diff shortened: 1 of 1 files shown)


fdehau:
Thanks for contributing. While I understand the intention of preserving the backward compatibility, I think that the API would be more intuitive if the header became optional, `None` by default and `Table::header` would set it. What do you think ?

It would be nice if you could add some tests for this new behavior too.

capgelka:
There are 3 places in the master branch to set a header

- `default` (No problems here it can be both `None` or `Some(H::Default)`. It looks like people don't use it preferring the `new` method. So changing its behavior should not be a problem.
- `new`. This is the most popular place to set a header. So changing its signature to `pub fn new(header: H, rows: R) -> Table<'a, H, R>` setting header field to `None` implicitly would break the API. Changing type to `Option<H>` would also break it and has another drawback, Rust cannot infer type for `None` in this case, and one should provide type annotation in this case
- `header`. As I see in apps mentioned in Readme people creating a table with `new` method, so it is not popular. However, changing type to optional we again have a problem with type inference for `None`.

So the `headerless` method is useful to solve the problem with `None`, and there is no need to change the `header` signature.
However, changing the default value to None is really more intuitive and we don't need a new method in this case.

We can also change `header` (and new) to use the `Into` trait.
```
   pub fn header<II, U>(mut self, header: II) -> Table<'a, H, R>
    where
        II: Into<Option<U>>,
        U: IntoIterator<Item = H::Item, IntoIter = H>
    {
        self.header = header.into().map(|h| h.into_iter());
        self
    }
```

Probably I am doing something wrong here, but this works ok for args without optional wrapping, and I got cannot infer type error for both `Some` and `None`, so I am not sure this if it a good idea.


So it looks like change default value is enough to allow having a "headerless table", the question is should it be another more explicit method to set header to None or not?
If there is a way to use None with type inference, of course, it's better then additional headerless function. But making user set None type directly doesn't look like a better option. Unfortunately, I am not experienced enough to understand if it is possible or not to avoid it.

Is changing default value and saving headerless function the best solution? If so I will change the `default` and add tests.
 But if you have an idea of how to fix the problem with type inference I prefer first to try it. Of course, it's better to allow None explicitly, especially with the `Into` trait.

fdehau:
I would advocate for a breaking change if it improves the API as long as it is correctly documented and easy to migrate, which is the case at first view. I think changing header to be None by default, removing header from `Table::new` and having `Table::header` set the header to `Some(smth)` is probably the thing that is the most consistent with other widgets if we want to support headerless tables. What do you think ?

ivan-gulakov:
@fdehau hi.

We tried to change the API the way you wrote about it, but unfortunately Rust can't infer types like my colleague wrote above.

In addition, you cannot put an `Unsized `type inside `Option`, which is, in fact, `Display`. (Although the Rust devs want to improve this soon)

Maybe, it's better to allow set header to `None ` by `headerless `option. Also, this way we can ensure, that entries inside `Iterator`, passed to header will be `Display `type.

P.S. In such case the PR is ready.

fdehau:

Issue #239, opened 2020-03-06
Text Style Background should affect whole Row/Cell in List and Table


Ynscription:
When using Text::styled in a List or Table, the background color only affects the text. I believe it should affect the whole Row/Cell in which the text is in.

**To Reproduce**
```
let splits_list = splits.iter().map(|(txt)| {
    Text::styled(Cow::from(txt), 
        Style::default()
            .bg(Color::Red))
});
let mut splits_list = List::new(splits_list)
    .style(normal_style1);
f.render(&mut splits_list, chunks[0]);
``` 

**Screenshots**

![Tui List BG](https://user-images.githubusercontent.com/6177465/76576577-c2a13800-64c2-11ea-96ec-ac654266d995.png)

**Desktop:**
 - OS: Tested on [Linux Mint Cinnamon and Windows 10]
 - Terminal Emulator [Gnome Terminal, Power Shell and Cmd]
 - Font [Monospace and Consolas]
 - Crate version [0.8]
 - Backend [crossterm]

fdehau:
I'm not sure how I feel about this. While I see why one could expect this behavior, the proposed change is kind of implicit. What do you think if instead we introduced a wrapping `Item` struct where this styling can be applied ? Something like:

```rust
struct Item<'a> {
    content: Text<'a>,
    style: Style
}
```
where in your case:
```rust
let item = Item {
    content: Text::styled(txt, Style::default().fg(Color::Green)),
    style: Style::default().bg(Color::Green)
};
```

Ynscription:
That sounds like a great solution to keep the previous functionality and add the expected behaviour I described.

Extrawurst:
Having the same issue! It would be great to have a solution to it.

BTW. this project is awesome! :)

pebbleKite:
@fdehau @Ynscription 
1. What would you say the best interaction of `struct Item` with `selected`? Should the selected styling affect the wrapping layer or the content only?
2. In the table case, should the wrapping style of `Item` be applied to the entire row, or each cell be an instance of `Item` (instead of the current `D where D: Display`)

fdehau:
@pblkt 
1. I think it's reasonable to expect that a user would be able to customize the style of both the `Item` and its content when it is selected. This may imply a breaking change in the way we expose this setting.
2. It's a bit early to apply the same logic to table but given that the table is currently row oriented it would make sense to have a more complex `Row` struct instead of a introducing an `Item`.

teohhanhui:
[...]

PR #317, opened 2020-06-07
Implement style merging for highlight styles

Rantanen:
Previously the highlight style (even if not present) overrode the row styles for a table and list:

image

With this change, if there is no highlight style defined, the selected item follows the style of the row:

image

When a highlight style is defined, it is _merged_ with the data style:

image

---------

This is now implemented on top of `StyleDiff`. The `highlight_style` and `header_style` are accepted through that. The basic `style` is still a normal `Style` as this should be the lowest priority style so there's nothing to patch that on top of.

I also wish I could have added `DiffStyled` variants to `Text` and `Row`, but that would have been a compatibility break. Maybe it should be considered for the next semver-major release?

There's also some overall improvements here in how styles get applied between highlight symbols, etc.


fdehau:
Thanks for your contribution. The highlighting behavior has been certainly overlooked until now. Given the ongoing discussion in #239, I had already started some experimentation on the styling issues of List and reached the same conclusion as you that we might want to have a style incrementally applied instead of overridden. I've cleanup this work, taking your ideas into account in #318. I think that as you suggested using a `StyleDiff` in a backward-compatible way (there is a `From<Style>` impl) for `highlight_style` would be a much better idea as it would not require an implicit merge logic as you introduced in this first version. Let me know what you think. If you are ok the plan would then be to merge #318 and let you rework your PR upon it.

Rantanen:
That sounds great! I was somewhat reluctant to push completely new type out of the blue, but given #318 exists, I feel the use of `StyleDiff` makes more sense for the API for this PR.

I'll rework this implementation once that PR land.

fdehau:
I've merged #318 so the road is clear ;).

Rantanen:
This needs some backwards compatibility tests. I patched my project on top of this, expecting the behavior to revert to the original (until I fixed my project to use the new `_diff` methods), but didn't see any change.

I'll add some tests that verify the old behavior doesn't change and see if there's something that changes it with the implementation.

Rantanen:
There are two changes that affect backwards compatibility:

- The way `Row::Data` is styled has changed, which is to say it is now styled.
- The parts that are affected by `Row::StyledData` differ slightly to make `Table` consistent with `List`.

In the past `Row::Data` was always unstyled (or styled with `Style::default()`), now `Table::style` is applied to that data instead. I'd consier the previous behavior a bug, essentially `Table::style().fg` was not used for anything and `Table::style().bg` was used only for empty spaces.

`Row::StyledData` style applied to also the blank spaces where `highlight_symbol` would be. This wasn't visible for foreground color, but if the `Row::StyledData` applieda  background color, this extended also to the highlight symbol column. This differs from `List`, which always had the highlight symbol color either as `List::style()` for blank symbols or `List::highlight_style()` if the row was highlighted.

The diff to backwards compatibility changes is:
https://github.com/fdehau/tui-rs/pull/317/commits/1614bf269c564db443ef2c4488113046ed74e635#diff-538c05d01c0ce71eadc533e09c6f7194R540-R592

Furthermore for backwards compatibility reasons the _default_ for `style`/`highlight_style`/etc. is now `StyleDiff::from(Style::default())`, which is to say those styles will always fully override the previous styles.

List example (with `higlight_style` removed)

image

The behavior here is maintained between versions. Adding `hilight_style_diff(StyleDiff::default())` removes the backwards compatibility default style, resulting in:

image

Table example (with `style(Style::default().bg(Color::DarkGray))`)

Before the changes:
image

After the changes:
image

(Note the rows are all `StyledRow` so they still maintain the black background coming from `Style::default()`)

Adding `header_style_diff(StyleDiff::default())` removes the otherwise-overriding default style from headers and changing `highlight_style` to `highlight_style_diff` gives the following result:

image

src/widgets/list.rs
```diff
@@ -120,7 +113,12 @@ where
     }
 
     pub fn highlight_style(mut self, highlight_style: Style) -> List<'b, L> {
-        self.highlight_style = highlight_style;
+        self.highlight_style_diff = highlight_style.into();
+        self
+    }
+
+    pub fn highlight_style_diff(mut self, highlight_style: StyleDiff) -> List<'b, L> {
```
fdehau, line 120: I don't think that we need a second method here:

```rust
pub fn highlight_style<S>(mut self, highlight_style: S) -> List<'b, L> where S: Into<StyleDiff> {
}
```
The documentation should probably includes an example for both cases: `Style` and `StyleDiff` and explain how they differ. What do you think ?

fdehau:
I think your changes are very sensible but I would not worry too much about backward compatibility in this case as you pointed that the previous behavior was not consistent. If using `StyleDiff::default()` instead of `StyleDiff::from(Style::default())` provides better a out of the box result then let's change it. It will be especially nicer to newcomers and makes the API more understandable.

src/widgets/list.rs
fdehau: Actually reading your PR again, this might be confusing to some users. I think we will keep your new `diff` variants and deprecate the old ones.

Rantanen:
I'll deal with this next week. Currently spending midsummer at a summer house so not in a great position to make the changes at the moment.

I do agree that not bothering with the full backwards compatibility would result in less surprising user experience.

Rantanen:
Got back from the vacation to find my development server with one less working hard drive so that delayed this a bit.

Made the change to default to `StyleDiff::default()` (ie. empty `StyleDiff`) instead of the previous `Style::default()` (ie. style that resets everything).

This applies to list highlight style as well as table header and highlight style. So now the default behavior for a highlighted list/table row is to inherit the style from the row itself instead of using an empty `Style` that would reset the highlighted row/header.

Rantanen:
Based on the review comment that indicated the non-`StyleDiff` method should be deprecated (Which I agree with); I've added `#[deprecated]` attributes on the old methods and cleaned up the examples to use the new `_diff` methods instead. The version for `#[deprecated(since = "...")]` is `0.10.0`; That should be changed once the version in which this is going out is known.

Also saw that `widgets::Tabs` had a `highlight_style` as well so I changed that to use the `StyleDiff` for consistency as well.

fdehau:

PR #318, opened 2020-06-08
Style diff

fdehau:
Add `StyleDiff` that may be used in a context where we want to use an existing `Style` as a base for an other one. The more telling example at the moment is the highlighting behavior upon selection in several widgets. In this case, users may want to change only a single property of the current item style instead of overriding it completely.


extrawurst:
@fdehau nice that's pretty handy!👍

PR #319, opened 2020-06-08
Added margins to tabs widget

### LINKED ISSUE #299: Inconsistent margin spacing in widgets
**Describe the bug**
Widgets have inconsistent spacing on the margins in the blocks. Some widgets, for example the `Tabs`, have one column of spacing, while others like `Table` have no spacing. This just doesn't look right.

**Expected behavior**
All widgets should have consistent spacing (one or none) and optionally should be defined by the user for greatest flexibility.

**Screenshots**
Screen Shot 2020-05-22 at 2 24 10 PM

thorlucas:
Closes #299 

Margins can now be added to the tabs widget by using the `Margin` struct, which was previously used for margins between blocks I believe. The [`tabs.rs` example was updated to reflect this](https://github.com/thorlucas/tui-rs/blob/af2f0d47b1e4d4a7ca1a177e1c10d4c9aa5fd53f/examples/tabs.rs#L56).

The `vertical` field sets the margins/padding on the top and bottom of the widget, while the `horizontal` on the left and right. The tabs widget will be trimmed if these margins would push the contents off the screen. For example, setting vertical margins to 1 on a 2-line-tall tabs widget would display nothing, as the two lines would be used as margins.

The margins can be set with a `.margin(...)` builder method on the tabs widget.

Builder methods were also added for the `Margin` struct to facilitate creation. The `Margin::default()` sets both to zero, as is the visual default for other widgets, while the `Margin::default().horizontal(x)` and `.vertical(x)` builder methods set the horizontal and vertical components. `.both(xy)` sets both vertical and horizontal components to the same number.

Several new tests were added to the `widgets_tabs.rs` to ensure that the widget respects horizontal and vertical margins in a variety of cases.

Note that margins will still need to be added to other widgets in able to behave consistently with the new style of adding margins (`.margin(...)` with a `Margin` struct). These will be submitted as separate PRs.

### DIFF
--- a/src/widgets/tabs.rs
+++ b/src/widgets/tabs.rs
@@ -118,38 +118,48 @@ where
 
         buf.set_background(tabs_area, self.style.bg);
 
-        let mut x = tabs_area.left();
+        let mut x = tabs_area.left() + self.margin; // adds left margin
+        let mut space_remaining: isize =
+            (tabs_area.right() - tabs_area.left()) as isize - (self.margin * 2) as isize;
+
         let titles_length = self.titles.len();
-        let divider_width = self.divider.width() as u16;
-        for (title, style, first_title, last_title) in
-            self.titles.iter().enumerate().map(|(i, t)| {
-                let lt = i + 1 == titles_length;
-                let ft = i == 0;
-                if i == self.selected {
-                    (t, self.highlight_style, ft, lt)
-                } else {
-                    (t, self.style, ft, lt)
-                }
-            })
-        {
-            if first_title {
-                // Adds left margin on first title
-                x += self.margin;
+        // divider actually requires a space before it, so we add one
+        let divider_width = self.divider.width() as u16 + 1;
+        for (title, style, last_title) in self.titles.iter().enumerate().map(|(i, t)| {
+            let lt = i + 1 == titles_length;
+            if i == self.selected {
+                (t, self.highlight_style, lt)
             } else {
-                // Adds one space between titles
-                x += 1;
+                (t, self.style, lt)
+            }
+        }) {
+            if space_remaining <= 0 {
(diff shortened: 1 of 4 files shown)


src/widgets/tabs.rs
```diff
@@ -121,17 +136,33 @@ where
                 (t, self.style, lt)
             }
         }) {
-            x += 1;
             if x >= tabs_area.right() {
                 break;
+            }
+
+            let mut space_remaining: isize = (tabs_area.right() as isize) - (x as isize);
```
fdehau, line 143: The code usually avoids dealing with signed integer to limit conversions. You can use saturing add and sub operations to avoid overflows.

thorlucas: I'm not sure I understand what you mean by this.

thorlucas:
Additionally, if this margin style works for you, please give me the go ahead to continue implementing this for all widgets.

src/widgets/tabs.rs
fdehau: No I meant that usually I try to keep using u16 or usize because you can't represent impossible states (e.g negative width) and as a tendency to panic more often which makes the debugging experience easier and makes you more aware of the edge cases.

fdehau:
As you offered to work on other widgets as well, I'm wondering if we should take this opportunity to rename `Margin` to `Padding` (as suggested sometime ago @teohhanhui) as it might make more sense to newcomers and people familiar with other UI libraries.

thorlucas:
Sure!

src/widgets/tabs.rs
thorlucas: Ah, alright I'll change that up. Thanks

Issue #320, opened 2020-06-09
Ability to combine `StyleDiff` into one


Rantanen:
Short version:

```rust
// Being able to patch multiple `StyleDiff` into one is what I wanted.
fn apply(s: Style, d1: StyleDiff, d2: StyleDiff) {
    let combined = d1.patch(d2);
    assert_eq!(
        s.patch(d1).patch(d2),
        s.patch(combined));
}

// This brought me to ponder the following situations

// User specifies `modifier`, I feel this should override any existing
// `add_modifier`/`remove_modifier`
//
// Currently any `add_modifier`/`remove_modifier` change are applied on top of this.
fn set_mod(s: Style, d: StyleDiff) {
    let d = d.modifier(Modifier::BOLD);
    assert_eq!(s.patch(d).modifier, Modifier::BOLD);
}

// User wants to `add_modifier`. This should come into effect even if original
// diff specified `modifier` or `remove_modifier`.
//
// Currently `remove_modifier` is applied after `add_modifier`.
fn add_mod(s: Style, d:StyleDiff) {
    let d = d.add_modifier(Modifier::BOLD);
    assert_eq!(s.patch(d).modifier & Modifier::BOLD, Modifier::BOLD);
}
```

(The implementation would involve resetting `add_modifier`/`sub_modifier` when `modifier` is set and doing `other_modifier.remove(this_modifier)` when `this_modifier` is modified to ensure there's no overlap.)

.. Oh, and also I want to change the add/sub fields from `Option<Modifier>` into just `Modifier`. The empty bitflag field already has the equivalent effect of "Do not change anything" - there's no need for `None` to represent that.

-------------

Long version:

**Is your feature request related to a problem? Please describe.**
Given the fields of `StyleDiff` are private, I don't think there's any (easy?) way for user code to combine multiple `StyleDiff` into one. 

**Describe the solution you'd like**
Implement `fn patch(self, StyleDiff) -> StyleDiff` on `StyleDiff`

**Describe alternatives you've considered**
None

**Additional context**
Faced the need for this when looking into how base List/Table `style` should be combined with `highlight_style`.

I already kinda implemented this, but this brought up a slight refinement to how `StyleDiff` itself is implemented.

Currently the `add_modifier`/`sub_modifier` are `Option` types. Yet these being bitflags already kind of make them option types with empty values having no effect on the final style. While these are implemented with `Option`, combining styles together now results in code such as:

```rust
[...]

PR #321, opened 2020-06-09
Implement `patch` between two `StyleDiff`

Rantanen:
Almost had need for this during the highlight style implementation, but avoided it in the end since `style` remained `Style` instead of `StyleDiff`.

In any case, I figured implementing this acted as a good exercise for just what a specific `StyleDiff` should mean in regards to what it does when it's applied (especially in relation to what modifiers get applied, if a `StyleDiff` specifies all `modifier`, `add_modifier` and `sub_modifier` with various degrees of overlap).

The PR changes the existing semantics in this regard by making the order in which modifiers are defined important (`modifier` overrides existing add/remove details and `add`/`remove` build on top of the last `modifier` while avoiding overlap between each other).

Even if the `patch` functionality wasn't needed, I feel refining `StyleDiff` functionality now makes sense to avoid regrets once it can't be changed anymore due to backwards compatibility issues.

Implements #320 (which also has more background details on the actual changes).

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -190,12 +190,8 @@ impl Style {
         if let Some(m) = diff.modifier {
             self.modifier = m;
         }
-        if let Some(m) = diff.add_modifier {
-            self.modifier.insert(m)
-        }
-        if let Some(m) = diff.sub_modifier {
-            self.modifier.remove(m)
-        }
+        self.modifier.insert(diff.add_modifier);
+        self.modifier.remove(diff.sub_modifier);
         self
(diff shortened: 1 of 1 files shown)


fdehau:
I think that your `patch` implementation is more sensible than mine so I would happily take a PR to improve it. But I would rather keep this method only on `Style` for now until there is a real use case for it on `StyleDiff`.

fdehau:
I finally found a use case while working on #316 so merging this. Thanks !

PR #322, opened 2020-06-10
Update paragraph.rs

popey456963:
Adds a note about paragraphs in blocks.  This is after a bug we struggled with for half a day.


fdehau:
This behavior is shared by all widgets using a block. Would you consider this concept easier to approach if it was mentioned on the `Block` documentation directly ?

PR #325, opened 2020-06-17
Add diskonaut to "apps using tui"

imsnif:
Hey @fdehau,

I created another app using your cool library. It was really fun implementing a treemap using the infrastructure you created.

Thanks again for this library!


fdehau:
I'm always happy to see what people end up building with the help of this crate. The treemap layout is very impressive indeed !

PR #327, opened 2020-06-23
Preserve Indentation of Unwrapped Lines

TheLostLambda:
Thanks for the wonderful library! :heart: 

## Issue

I ran into some unexpected behaviour recently when I was attempting to print a paragraph with some indentation in wrapping mode – TUI was stripping the leading whitespace from my lines.

I've since discovered that this is an intentional behaviour, but that leaves me **unable** to print a paragraph like this:
```
MW Word of the Day                                                            
*verb*                                                                        
   1) to diffuse through or penetrate something                               
   2) to spread or diffuse through                                            
   3) to pass through the pores or interstices of
```

## Changes

In this PR, I simply disabled this behaviour, as I feel it's more predicable.

The above paragraph now wraps as follows:
```
MW Word of the Day            
*verb*                        
   1) to diffuse through or   
penetrate something           
   2) to spread or diffuse    
through                       
   3) to pass through the     
pores or interstices of
```

### Justification

Slightly funky? Perhaps, but much more predicable and arguably clearer than no indentation whatsoever.

Comparing to the current master TUI branch:
```
MW Word of the Day            
*verb*                        
1) to diffuse through or      
penetrate something           
2) to spread or diffuse       
through                       
3) to pass through the pores  
or interstices of 
```

The lack of indentation makes it hard to tell that those are distinct "bullet points". **The user can always trim strings in their application, but can't re-add the spacing TUI strips.**

## Tests
I adapted one of the leading-space tests to test indentation is preserved, but the other two I removed. If you are curious, the output of the removed tests was:
```
---- widgets::reflow::test::line_composer_char_plus_lots_of_spaces stdout ----
thread 'widgets::reflow::test::line_composer_char_plus_lots_of_spaces' panicked at 'assertion failed: `(left == right)`
  left: `["a", "                    ", "                    ", "       "]`,
 right: `["a", ""]`', src/widgets/reflow.rs:401:9

---- widgets::reflow::test::line_composer_lots_of_spaces stdout ----
thread 'widgets::reflow::test::line_composer_lots_of_spaces' panicked at 'assertion failed: `(left == right)`
  left: `["                    ", "                    ", "                    ", "      "]`,
 right: `[""]`', src/widgets/reflow.rs:385:9
```
Are a load of lines filled by nothing but spaces useful? Probably not, but at least this gives the user the choice to keep or remove them. This new behaviour is also somewhat more in line with the `LineTruncator` which leaves the trailing whitespace.

### DIFF
--- a/src/widgets/reflow.rs
+++ b/src/widgets/reflow.rs
@@ -20,18 +20,21 @@ pub struct WordWrapper<'a, 'b> {
     max_line_width: u16,
     current_line: Vec<Styled<'a>>,
     next_line: Vec<Styled<'a>>,
+    trim: bool,
 }
 
 impl<'a, 'b> WordWrapper<'a, 'b> {
     pub fn new(
         symbols: &'b mut dyn Iterator<Item = Styled<'a>>,
         max_line_width: u16,
+        trim: bool,
     ) -> WordWrapper<'a, 'b> {
         WordWrapper {
             symbols,
             max_line_width,
             current_line: vec![],
             next_line: vec![],
+            trim,
         }
     }
 }
@@ -60,8 +63,8 @@ impl<'a, 'b> LineComposer<'a> for WordWrapper<'a, 'b> {
 
             // Ignore characters wider that the total max width.
             if symbol.width() as u16 > self.max_line_width
-                // Skip leading whitespace.
-                || symbol_whitespace && symbol != "\n" && current_line_width == 0
+                // Skip leading whitespace when trim is enabled.
+                || self.trim && symbol_whitespace && symbol != "\n" && current_line_width == 0
             {
                 continue;
             }
@@ -234,6 +237,7 @@ mod test {
 
     enum Composer {
         WordWrapper,
+        WordWrapperNoTrim,
         LineTruncator,
     }
 
@@ -241,7 +245,10 @@ mod test {
         let style = Default::default();
         let mut styled = UnicodeSegmentation::graphemes(text, true).map(|g| Styled(g, style));
         let mut composer: Box<dyn LineComposer> = match which {
-            Composer::WordWrapper => Box::new(WordWrapper::new(&mut styled, text_area_width)),
+            Composer::WordWrapper => Box::new(WordWrapper::new(&mut styled, text_area_width, true)),
+            Composer::WordWrapperNoTrim => {
+                Box::new(WordWrapper::new(&mut styled, text_area_width, false))
+            }
             Composer::LineTruncator => Box::new(LineTruncator::new(&mut styled, text_area_width)),
         };
         let mut lines = vec![];
@@ -480,5 +487,24 @@ mod test {
         let text_space = text.replace("\u{00a0}", " ");
         let (word_wrapper_space, _) = run_composer(Composer::WordWrapper, &text_space, width);
         assert_eq!(word_wrapper_space, vec!["AAAAAAAAAAAAAAA AAAA", "AAA",]);
+    }
+
+    #[test]
+    fn line_composer_word_wrapper_preserve_indentation() {
+        let width = 20;
+        let text = "AAAAAAAAAAAAAAAAAAAA    AAA";
+        let (word_wrapper, _) = run_composer(Composer::WordWrapperNoTrim, text, width);
+        assert_eq!(word_wrapper, vec!["AAAAAAAAAAAAAAAAAAAA", "   AAA",]);
+    }
+
+    #[test]
+    fn line_composer_word_wrapper_preserve_indentation_with_wrap() {
+        let width = 10;
+        let text = "AAA AAA AAAAA AA AAAAAA\n B\n  C\n   D";
+        let (word_wrapper, _) = run_composer(Composer::WordWrapperNoTrim, text, width);
+        assert_eq!(
+            word_wrapper,
+            vec!["AAA AAA", "AAAAA AA", "AAAAAA", " B", "  C", "   D"]
+        );
     }
 }
--- a/src/widgets/reflow.rs
+++ b/src/widgets/reflow.rs
@@ -20,6 +20,7 @@ pub struct WordWrapper<'a, 'b> {
     max_line_width: u16,
     current_line: Vec<Styled<'a>>,
     next_line: Vec<Styled<'a>>,
+    /// Removes the leading whitespace from lines
     trim: bool,
 }
 
@@ -236,8 +237,7 @@ mod test {
     use unicode_segmentation::UnicodeSegmentation;
 
     enum Composer {
-        WordWrapper,
-        WordWrapperNoTrim,
+        WordWrapper(bool),
         LineTruncator,
     }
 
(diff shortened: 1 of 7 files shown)


fdehau:
Instead of removing the existing the existing behavior that might be still useful to some users, have you considered exposing this as an option on the composers ?

TheLostLambda:
That's certainly something I could look into! Are you thinking some sort of builder method like `.wrap()`? Perhaps something like `.trim()`? If we do something like that, should the default behaviour be removing the whitespace or keeping it?

I'd be happy the try to find a middle-ground :)

fdehau:
Yeah adding a trim option is good idea. To preserve backward compatibility for now, I would suggest that `trim` be true by default, unless we see a lot of reports of people confused by this.

TheLostLambda:
Sounds good! I'll add that to my schedule for tomorrow and will try to come up with something backwards compatible and clean :) Thanks for the pointers today!

TheLostLambda:
Hey @fdehau ! I've given this some testing and it works for my application without any changes in the default behaviour – I just call  `.trim_wrapped(false)` on my `Paragraph` widget and everything works! I've also added a couple of new tests for this indentation preservation and all of the tests for the previous behaviour are still in there :)

Let me know what you think!

src/widgets/reflow.rs
```diff
@@ -196,14 +199,18 @@ mod test {
 
     enum Composer {
         WordWrapper,
+        WordWrapperNoTrim,
```
fdehau, line 202: We could maybe have `WordWrapper { trim: bool }` instead ?

src/widgets/reflow.rs
```diff
@@ -443,4 +450,23 @@ mod test {
...
+        let width = 20;
+        let text = "AAAAAAAAAAAAAAAAAAAA    AAA";
+        let (word_wrapper, _) = run_composer(Composer::WordWrapperNoTrim, text, width);
+        assert_eq!(word_wrapper, vec!["AAAAAAAAAAAAAAAAAAAA", "   AAA",]);
+    }
+
+    #[test]
+    fn line_composer_word_wrapper_preserve_indentation_with_wrap() {
+        let width = 10;
+        let text = "AAA AAA AAAAA AA AAAAAA\n B\n  C\n   D";
```
fdehau, line 465: Shouldn't we have a case of a line leading with white-spaces but whose length is greater than the width ?

src/widgets/reflow.rs
```diff
@@ -19,18 +19,21 @@ pub struct WordWrapper<'a, 'b> {
     max_line_width: u16,
     current_line: Vec<Styled<'a>>,
     next_line: Vec<Styled<'a>>,
+    trim: bool,
```
fdehau, line 22: Could you add a doc comment on what this option do ?

src/widgets/reflow.rs
TheLostLambda: For sure! I was originally looking for the minimal number of changes to the other tests, but this is probably the cleaner approach :)

src/widgets/reflow.rs
TheLostLambda: A very good catch! I'll add one! Sorry for missing that the first time around.

src/widgets/reflow.rs
TheLostLambda: I've gone with an `enum` variant for the moment, but let me know if you think this would be better done with `structs`!

fdehau: I agree with keeping an enum variant here but we could maybe go for an anonymous struct so it is more obvious what the boolean flag is for, e.g.:
```rust
enum Composer {
  WordWrapper { trim: bool },
  LineTruncator
}
```

TheLostLambda: That's an excellent idea, sorry for missing that! I forgot the you could have Enum structs like that! I'll make that change now :)

fdehau:
The changes and the tests look good. I'm just wondering if we could improve the API a bit further, I worry that having a distinct method on the widget will hurt the discoverability of this behavior. What do you think if we had the following:

```rust
struct Wrap {
  pub trim: bool
}

struct Paragraph {
   wrap: Option<Wrap>,
}

impl Paragraph {
  fn wrap(mut self, wrap: Wrap) -> Paragraph {
    self.wrap = Some(wrap);
    self
  }
}
```
so that it used like this:
```rust
// No wrapping
let p = Paragraph::default();
// Wrap but not trim
let p = Paragraph::default().wrap(Wrap { trim: false });
// Wrap and trim
let p = Paragraph::default().wrap(Wrap { trim: true });
```
My motivation is that when a user want wrapping in a paragraph, they are already aware of this option and have to look up what it does.

We could also add your bullet points example in the documentation as it helps visualize the difference very quickly.

TheLostLambda:
Hi @fdehau !

I've thought over that proposed API and while I agree it would make the trim option much more visible, it seems to come at the cost of some breaking changes and ultimately more verbose code. While I obviously need this trim option, it seems like plenty of applications built on tui get by fine without needed to think about trimming on wrap. I think that all of those cases are still best suited by the default behaviour and might not need breaking changes.

Though I personally might prefer that the builder-style API is used, I'm open to giving this alternative struct-based API a shot if that's what you think is best.

We can certainly still include that documentation showing off the trim option with the example in this version as well; that might draw some additional attention to the feature.

Let me know what you think!

fdehau:
> it seems to come at the cost of some breaking changes

This change will be included in a release that is likely to include a lot of breaking changes (though easy to fix). I wouldn't worry so much about that in this discussion. We could add a `From<bool>` implementation with a deprecation notice if we really want to avoid this. Plus I always include a migration guide in the changelog for each of those breaking changes.

> and ultimately more verbose code

I don't really buy this from the end-user perspective:
```rust
.wrap(Wrap { trim : true })
```
vs
```rust
.wrap(true)
.trim_wrapped(true)
```
is about the same amount of characters and the both solutions are explicit enough.

> While I obviously need this trim option, it seems like plenty of applications built on tui get by fine without needed to think about trimming on wrap

But that was exactly my point. I want to be 100% that people have seen this option and made a conscious choice. If the wrapping result is not the one they expected they already know that they have at least one option to tweak. Beside if the documentation is clear enough about the differences of the two behaviors, that will certainly make the library easier to use and reduce the chance of people asking questions which is always good from a maintainer point of view :).

> We can certainly still include that documentation showing off the trim option with the example in this version as well; that might draw some additional attention to the feature.

If you could include that in code examples of the `wrap` method it would be awesome.

TheLostLambda:
Hi @fdehau ! You bring up some very good points, and I'd be happy either way, so I think I'll re-implement it tomorrow using the API you've proposed! I'll try to find a place for that example as well :smile:

TheLostLambda:
Hey @fdehau !

I've moved to the API you proposed and have added some documentation for the `Wrap` struct, showing the effect it has on indented text.

Let me know what you think!

fdehau:
This looks good. Some conflicts to resolve and we will be good to go =).

TheLostLambda:
Haha, darn! I almost snuck ahead of that other PR. Things should be rebased now :)

PR #329, opened 2020-06-25
Add scroll horizontally support for paragraph

xiaopengli89:
Now Paragraph can scroll horizontally
```rust
let p = Paragraph::new(items)
    .block(
        Block::default()
    )
    .scroll((0, 7));
```

```
+------------------------------+
|hello world                   |
+------------------------------+
```

```
+------------------------------+
|orld                          |
+------------------------------+
```

### DIFF
--- a/src/widgets/paragraph.rs
+++ b/src/widgets/paragraph.rs
@@ -136,7 +136,11 @@ where
         let mut line_composer: Box<dyn LineComposer> = if self.wrapping {
             Box::new(WordWrapper::new(&mut styled, text_area.width))
         } else {
-            Box::new(LineTruncator::new(&mut styled, text_area.width, self.scroll.1))
+            Box::new(LineTruncator::new(
+                &mut styled,
+                text_area.width,
+                self.scroll.1,
+            ))
         };
         let mut y = 0;
         while let Some((current_line, current_line_width)) = line_composer.next_line() {
@@ -144,11 +148,7 @@ where
                 let mut x = get_line_offset(current_line_width, text_area.width, self.alignment);
                 for Styled(symbol, style) in current_line {
                     buf.get_mut(text_area.left() + x, text_area.top() + y - self.scroll.0)
-                        .set_symbol(if symbol.is_empty() {
-                            " "
-                        } else {
-                            symbol
-                        })
+                        .set_symbol(if symbol.is_empty() { " " } else { symbol })
                         .set_style(*style);
                     x += symbol.width() as u16;
                 }
--- a/src/widgets/paragraph.rs
+++ b/src/widgets/paragraph.rs
@@ -139,7 +139,10 @@ where
             Box::new(LineTruncator::new(
                 &mut styled,
                 text_area.width,
-                self.scroll.1,
+                match self.alignment {
+                    Alignment::Left => self.scroll.1, // only support Alignment::Left
+                    _ => 0,
+                },
             ))
         };
         let mut y = 0;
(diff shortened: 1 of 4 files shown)


fdehau:
Thanks that looks like a nice addition. Could you add some tests to cover this new behavior ?

xiaopengli89:
I've add some tests

src/widgets/paragraph.rs
```diff
@@ -136,21 +136,28 @@ where
...
         let mut y = 0;
         while let Some((current_line, current_line_width)) = line_composer.next_line() {
-            if y >= self.scroll {
+            if y >= self.scroll.0 {
                 let mut x = get_line_offset(current_line_width, text_area.width, self.alignment);
                 for Styled(symbol, style) in current_line {
-                    buf.get_mut(text_area.left() + x, text_area.top() + y - self.scroll)
-                        .set_symbol(symbol)
+                    buf.get_mut(text_area.left() + x, text_area.top() + y - self.scroll.0)
+                        .set_symbol(if symbol.is_empty() { " " } else { symbol })
```
fdehau, line 154: Why was that change required ?

src/widgets/reflow.rs
```diff
@@ -124,16 +125,19 @@ pub struct LineTruncator<'a, 'b> {
     symbols: &'b mut dyn Iterator<Item = Styled<'a>>,
     max_line_width: u16,
     current_line: Vec<Styled<'a>>,
+    cut: u16,
```
fdehau, line 128: I'm not a big fan of the name. What do think of `horizontal_offset` ? Could you also add a doc comment to explain how is this property will be used when the `LineTrucator` process lines ?

src/widgets/reflow.rs
```diff
@@ -124,16 +125,19 @@ pub struct LineTruncator<'a, 'b> {
...
 }
 
 impl<'a, 'b> LineTruncator<'a, 'b> {
     pub fn new(
         symbols: &'b mut dyn Iterator<Item = Styled<'a>>,
         max_line_width: u16,
+        cut: u16,
```
fdehau, line 135: Given that it is not a required property, could we default to `0` and have a different method to set it ?

src/widgets/reflow.rs
```diff
@@ -189,6 +207,20 @@ impl<'a, 'b> LineComposer<'a> for LineTruncator<'a, 'b> {
     }
 }
 
+fn trim_width(src: &str, mut wid: usize) -> &str {
```
fdehau, line 210: Even though the function name is quite explicit, could you add a small doc comment on what this function does ?

src/widgets/paragraph.rs
xiaopengli89: Thanks for your reviews. If the symbol is empty, the last char which rendered last time will leave on the line. May be there are other better solutions.

src/widgets/reflow.rs
xiaopengli89: I agree with you, and will add a doc to explain.

src/widgets/reflow.rs
xiaopengli89: That's a good idea.

src/widgets/reflow.rs
xiaopengli89: Okay.

src/widgets/paragraph.rs
fdehau: No that seems right but could you add a comment to explain this. It will make the code easier to follow next time somebody has to read this part of the code :).

xiaopengli89: Agree with you.

fdehau:
Thanks a lot ! That's a nice addition.

Issue #339, opened 2020-07-17
Version 0.10: Spans can't contain line-breaks


TheLostLambda:
**Describe the bug**
I understand that multiple lines in a paragraph are intended to be represented using `Text`, but I seem unable to apply a span over several lines which I think is still very useful.

**To Reproduce**
Using the current development version from Git, I'm ultimately building the following:

```rust
[src/console/render.rs:58] vec![Spans :: from(text), Spans :: from("This is on the second line...")] = [
    Spans(
        [
            Span {
                content: "To Kill a Mockingbird\n",
                style: Style {
                    fg: None,
                    bg: None,
                    add_modifier: ITALIC,
                    sub_modifier: (empty),
                },
            },
            Span {
                content: "*noun*\n   1) outfit, equipment, attire\n\n",
                style: Style {
                    fg: None,
                    bg: None,
                    add_modifier: BOLD,
                    sub_modifier: (empty),
                },
            },
            Span {
                content: "1) obstreperous\n",
                style: Style {
                    fg: None,
                    bg: None,
                    add_modifier: (empty),
                    sub_modifier: (empty),
                },
            },
            Span {
                content: "2) invective\n",
                style: Style {
                    fg: None,
                    bg: None,
                    add_modifier: (empty),
                    sub_modifier: (empty),
                },
            },
            Span {
                content: "3) habiliment\n",
                style: Style {
                    fg: None,
                    bg: None,
                    add_modifier: (empty),
                    sub_modifier: (empty),
                },
            },
            Span {
                content: "4) furore\n",
                style: Style {
                    fg: None,
                    bg: None,
                    add_modifier: (empty),
                    sub_modifier: (empty),
                },
            },
        ],
    ),
    Spans(
        [
            Span {
                content: "This is on the second line...",
                style: Style {
                    fg: None,
                    bg: None,
                    add_modifier: (empty),
                    sub_modifier: (empty),
                },
            },
        ],
    ),
]

```

Which results in the following (note that all of my `\n`'s were ignored):

```
To Kill a Mockingbird*noun*   1) outfit, equipment, attire1) obstreperous2)   
invective3) habiliment4) furore                                               
This is on the second line...      
```

**Expected behavior**

I'm really looking for what I had before the text refactoring:

```
To Kill a Mockingbird
*noun*
   1) outfit, equipment, attire

1) obstreperous
2) invective
3) habiliment
4) furore                                               
This is on the second line...      
```

**Desktop:**
 - OS: Linux
[...]

PR #341, opened 2020-07-20
Don't destroy all newlines in Spans

TheLostLambda:
This is the one-line fix to #339, but it was obviously there intentionally, so I'm open to discussing the merits of forcing the `Text` abstraction as the way to do multiple lines.

It's just that the current version makes applying one style to several lines much more difficult than it should be and makes double newlines extra messy.

Additionally, I've not tested it on Windows, but how does this newline filtering work on DOS files with the \r\n line ending?

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -172,11 +172,10 @@ impl<'a> Span<'a> {
         &'a self,
         base_style: Style,
     ) -> impl Iterator<Item = StyledGrapheme<'a>> {
-        UnicodeSegmentation::graphemes(self.content.as_ref(), true)
-            .map(move |g| StyledGrapheme {
-                symbol: g,
-                style: base_style.patch(self.style),
-            })
+        UnicodeSegmentation::graphemes(self.content.as_ref(), true).map(move |g| StyledGrapheme {
+            symbol: g,
+            style: base_style.patch(self.style),
+        })
     }
 }
 
--- a/src/text.rs
+++ b/src/text.rs
@@ -177,7 +177,6 @@ impl<'a> Span<'a> {
                 symbol: g,
                 style: base_style.patch(self.style),
             })
-            .filter(|s| s.symbol != "\n")
     }
 }
 


Byron:
I also fell for this when upgrading [`dua`](https://github.com/Byron/dua-cli) to v0.10. Besides the surprising behaviour, it's unclear what is supposed to be done instead in order to get an empty line. I found myself adding `Spans::from("")` to get newlines, which now is as expensive as allocating a `vec` with one element. Allowing to keep newlines would help with that, too.

fdehau:
@TheLostLambda 

> This is the one-line fix to #339, but it was obviously there intentionally

Yes you are right, I've put this there to enforce the contract that these three new text primitives are meant to expose. Removing this will produce unexpected results if new lines are used in the context of `List` (the only other places where `Text` is used for now).

> It's just that the current version makes applying one style to several lines much more difficult than it should be and makes double newlines extra messy.

I'm open to a discussion (which should probably happen on the issue for visibility) about but I need detailed examples of what are the pain points. My first impression is that the issue is not the design of the primitives themselves but more the lack of flexibility in how you can build them. I've thought about a builder with a style stack and maybe macros during the implementation but I haven't found much time to give it more thoughts.

> Additionally, I've not tested it on Windows, but how does this newline filtering work on DOS files with the \r\n line ending?

That's a good point. I'll probably need to revisit the behavior on Windows.

@Byron 

> Besides the surprising behavior, it's unclear what is supposed to be done instead in order to get an empty line. I found myself adding Spans::from("") to get newlines

You shouldn't have to add anything to get new lines, each line is simply one item of the `Vec` that can be used to build a `Text` and you can get empty lines with `Spans::from("")`. I suspect that I'm missing something but without examples it will be hard for me to get an understanding of what you are struggling with.

> which now is as expensive as allocating a vec with one element

I wouldn't worry too much about this unless you have numbers as there are way more expensive operations in `tui` at the moment. Beside you should be able to use `Spans::from(Vec::new())` to save an allocation (I could probably add `Spans::empty()` for this use case though).

TLDR: I would be more than happy to change things to make the text authoring more pleasant to work with but I need more feedback :).

dotdash:
First off, the new Text/Spans/Span approach allowed me to remove my custom multi-line List hack, thanks for that!

Ignoring the issue of nested vectors (which IMHO needs addressing as well, but I have not though of a comprehensive solution, yet), the main point for me is that things seem to be only "half way there". Converting a `&str` that contains newlines to `Spans´ using `From`/`Into` silently messes the result up, but directly converting it to a `Text` handles the newlines just fine. But then I cannot add styling, nor can I (ergonomically) make use of that functionality if I want anything else in that `Text` element.

For my use case, I guess I'd already be pretty happy with just two methods on `Text`, `append(&str)` and `append_styled(&str, Style)`, letting me just hand over a string and optionally a style, which is then broken into `Span`/`Spans`. The implementation might be a bit tricky as you would probably need some way to track the current state with regards to whether or not an `append` call starts a new `Spans` or just adds to the current one. I'm not sure what's the best approach there.

One note on allocations: about 95% of my `Spans` only contain a single span, which makes me feel like that's a valid case to optimize for, i.e. avoiding the `Vec` in the `Spans` for this case. Maybe by having an `enum`, maybe by representing `Text` as a single `Vec<Span>` to begin with, and storing line breaks in term of offsets into that `Vec`.

TheLostLambda:
Heya to anyone following this! I'm closing this in favour of #361 which fixes the same problem but in a more elegant way and without breaking the new text abstraction. Definitely let me know what you think of that new PR!

PR #349, opened 2020-08-02
feat(widgets/paragraph): improve scrolling

### LINKED ISSUE #89: How to scroll Paragraph when appending to the end?
I have a long multi-line string that I continuously append onto and I'd like `Paragraph` to append to the end. I'd then like `Paragraph` to scroll down so that no lines get hidden from the newly appended text. However, I can't know how much to scroll because `Paragraph` doesn't yield any information about its layouted text.

fdehau:
Add optional callback to `Wrap` to compute the scroll offsets given the wrapped lines. This let
users compute the offsets dynamically.

### DIFF
--- a/examples/demo/ui.rs
+++ b/examples/demo/ui.rs
@@ -274,7 +274,7 @@ where
             .fg(Color::Magenta)
             .add_modifier(Modifier::BOLD),
     ));
-    let paragraph = Paragraph::new(text).block(block).wrap(Wrap { trim: true });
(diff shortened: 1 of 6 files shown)


vinaychandra:
@fdehau are there any blockers for this PR?

deepu105:
@fdehau any plans on merging this? this would be quite nice to have so that workarounds with list can be avoided

help-t:
trying to bump this pr 😞

PR #350, opened 2020-08-06
feat: added type Block in canvas markers

Amjad50:
this allows for clearer colors than using Dot, especially when
decreasing the size of the terminal font in order to increase the
resolution of the canvas

the reason I think this can be useful, is here:

Before:
image

After:
image

These two images have the same resolution (font is the smallest in my terminal) but blocks will show more color than dots

I'm not sure why, but I'm trying to port my emulator UI to TUI, which is very cool

### DIFF
--- a/src/widgets/canvas/mod.rs
+++ b/src/widgets/canvas/mod.rs
@@ -111,26 +111,28 @@ impl Grid for BrailleGrid {
 }
 
 #[derive(Debug, Clone)]
-struct DotGrid {
+struct SingleCellGrid {
     width: u16,
     height: u16,
     cells: Vec<char>,
     colors: Vec<Color>,
+    cell_char: char,
 }
 
-impl DotGrid {
-    fn new(width: u16, height: u16) -> DotGrid {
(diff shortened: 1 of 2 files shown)


src/widgets/canvas/mod.rs
```diff
@@ -111,26 +111,28 @@ impl Grid for BrailleGrid {
 }
 
 #[derive(Debug, Clone)]
-struct DotGrid {
+struct SingleCellGrid {
```
fdehau, line 114: I would rather have it called `CharGrid`. WDYT ?

Amjad50: Yeah, I also like it

Amjad50:
I got another idea, what do you think of making it general, something like

```rust
Canvas::default().marker(symbols::Marker::Char('c'));
```

fdehau:
I've merged as is. I don't think we have a use case for generalizing the marker yet but that may definitely something to consider in the future. Thanks for your contribution :).

PR #351, opened 2020-08-14
add tickrs to apps using tui

tarkah:
Would be awesome if you could add my app to the list, thanks!

https://github.com/tarkah/tickrs


fdehau:
Nice !

PR #354, opened 2020-08-18
refresh current view

uuhan:
Seems this is no way to refresh current view. If there is another good way to do this, please correct me.

The case is, if there are some additional stdout output, the view will become very strange, like this:

issue

Can try with this demo file:

https://github.com/uuhan/tui-rs/blob/9337ad047a45a65f6320efc419178b71205de12e/examples/termion_demo.rs#L56

### DIFF
--- a/src/terminal.rs
+++ b/src/terminal.rs
@@ -311,4 +311,8 @@ where
     pub fn size(&self) -> io::Result<Rect> {
         self.backend.size()
     }
+
+    pub fn refresh(&mut self) -> io::Result<()> {
+        self.resize(self.size()?)
+    }
 }


fdehau:
If you write to stdout and say to tui it can write to the entire terminal screen, yes things are likely to end up messed up. What are you trying to achieve exactly here ?

uuhan:
> If you write to stdout and say to tui it can write to the entire terminal screen, yes things are likely to end up messed up. What are you trying to achieve exactly here ?

I need a way to refresh the view. I'am working on a quite complex terminal app and it relay's some other libraries and sometimes they may output some message to stdout. I can not control the behavior but I can offer a way to refresh the view, e.g. the user can press 'p' key to reset the view.

fdehau:
Would the new behavior of clear introduced in #360 fit your use case ?

fdehau:
Feel free to reopen if the suggested fix does not work :)

PR #361, opened 2020-08-23
Extend `Text` to be stylable and extendable

TheLostLambda:
This PR:

- Makes the `Text` constructors consistent with `Span` (adds `raw` and `styled`)
- Adds a `From<String>` implementation for `Text`
- Adds an `IntoIterator` implementation for `Text` (this is mainly for the `Extend` implementation, but it also means one can use `for` loops to iterate over the `Spans` of a `Text` object.
- Adds an `Extend` implementation for `Text`

This addresses #339 in a more thoughtful way than my last PR. To borrow a doctest from `Text` and show off the extension:

```rust
use tui::text::Text;
let mut text = Text::from("The first line\nThe second line");
assert_eq!(2, text.height());
text.extend(Text::raw("These are two\nmore lines!");
assert_eq!(4, text.height());
```  

There is obviously no documentation in this PR, as I don't know if this is an acceptable approach, but I'm happy to write that up if this is okay.

In the meantime, I'll be using these new extensions to re-implement some of my projects to test the ergonomics of this abstraction.

### DIFF
--- a/src/terminal.rs
+++ b/src/terminal.rs
@@ -31,7 +31,7 @@ impl Viewport {
 }
 
 #[derive(Debug, Clone, PartialEq)]
-/// Options to pass to [`Terminal::draw_with_options`]
+/// Options to pass to [`Terminal::with_options`]
 pub struct TerminalOptions {
     /// Viewport used to draw to the terminal
     pub viewport: Viewport,
--- a/src/text.rs
+++ b/src/text.rs
@@ -257,6 +257,28 @@ impl<'a> From<Spans<'a>> for String {
 
 /// A string split over multiple lines where each line is composed of several clusters, each with
 /// their own style.
+///
+/// A [`Text`], like a [`Span`], can be constructed using one of the many `From` implementations
+/// or via the [`Text::raw`] and [`Text::styled`] methods. Helpfully, [`Text`] also implements
+/// [`core::iter::Extend`] which enables the concatenation of several [`Text`] blocks.
+///
+/// ```rust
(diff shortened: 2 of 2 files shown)


TheLostLambda:
Hey @fdehau ! Sorry for the pester, but have you had the opportunity to look this PR over yet? The issue #339 contains some examples of the new interfaces.

fdehau:
Sorry @TheLostLambda for making you wait that long.

I've taken a look again and your changes seems to hit a sweet spot in terms of improvements to the developer experience with the new text API and retro-compatibility so I'll be happy to include them in a future release. The only thing that seems to be missing would be some tests (either directly in the docs or somewhere else) that showcase what these changes enable. Moreover, they may also prevent us from introducing an unwanted regression in the future.

TheLostLambda:
Lovely! No worries, this is a very busy project! I just wanted the all clear before putting together some tests :)

Would doc-tests and documentation (similar to the existing ones in that file) suffice?

fdehau:
> Would doc-tests and documentation (similar to the existing ones in that file) suffice?

Yes I think so. Ideally, the `Text` documentation should showcase all the ways you can build a `Text`. I believe you, as an app author,  may actually have better insights than me on how you wished it had been explained to you the first time :).

TheLostLambda:
Hi @fdehau ! I think I've decently documented things now and I also fixed a one-word link error in another file (`cargo doc` kept complaining, so I figured I'd update it).

Let me know what you think and if this looks good to you!

PR #362, opened 2020-08-23
Clean up some folds

TheLostLambda:
Unrelated to my other PR and is just stylistic, but I thought it was a little more Rust-y to use things like `max` and `sum`.

Pick and choose which changes you like I suppose (if any) :)


fdehau:
Thanks. It looks nicer indeed.

Issue #202, opened 2019-11-22
[FEATURE] Pie Chart


spenserblack:
I'd like to get your opinion on if it would be a good idea for tui to have a pie chart widget. If you think it's a good idea, I'd be happy to make a PR, but I understand if you think this feature shouldn't be added, since it probably *requires* color support to display properly, unlike other widgets which still look good without colors.

PR #364, opened 2020-08-24
Feature/pie chart

spenserblack:
For #202 

Most of this work was done back when this crate was on version *0.6.2* (:fearful:), so there probably needs to be some changes to this PR. Please review and let me know how this looks :bow:


spenserblack:
BTW, When you run the example, the pie chart may look more like an oval than a circle. This is due to terminal "rows" being more pixels tall than "columns" are wide.

src/widgets/piechart.rs
```diff
@@ -0,0 +1,109 @@
...
+            chart_area.x + chart_area.width / 2,
+            chart_area.y + chart_area.height / 2,
+        );
+        let radius = min(chart_area.height, chart_area.width) / 2;
+        if radius < 2 {
+            return;
+        }
+        let radius: f64 = (radius - 2).into();
+
+        let cell_count = buf.content.len();
```
fdehau, line 73: You are iterating over the content of the entire buffer here (i.e the full terminal) which seems pretty inefficient. Any reason to only consider the allocated area ?

src/widgets/piechart.rs
```diff
@@ -0,0 +1,109 @@
...
+    fn default() -> PieChart<'a> {
+        PieChart {
+            block: None,
+            angles: Vec::new(),
+        }
+    }
+}
+
+impl<'a> PieChart<'a> {
+    pub fn data(mut self, data: &'a [(f64, Color)]) -> PieChart<'a> {
```
fdehau, line 28: We may want to have some sort of labels or legend. Otherwise, how are we going to know which value is what ?

fdehau (changes requested):
Nice initiative :clap: I've left some initial comments. A screenshot and some tests would be nice too :).

> BTW, When you run the example, the pie chart may look more like an oval than a circle. This is due to terminal "rows" being more pixels tall than "columns" are wide.

Could we maybe use braille patterns to get a more accurate rendering ?

src/widgets/piechart.rs
spenserblack: My concern with adding a legend is that it will take up space inside the block. Perhaps a separate legend `Widget` could be added to let the user decide how much terminal space should be allocated?

spenserblack:
>Could we maybe use braille patterns to get a more accurate rendering?

I've done some testing with rendering circles in the terminal in a separate project, and it seems like a 8:15 width:height ratio for characters is typical, but I'll need to test in other terminals to confirm. As long as it's safe to assume this ratio, I should be able to get a nearly perfect circle.

src/widgets/piechart.rs
fdehau: I agree that a legend might be tricky to implement so we can leave it out of this PR but I still think that we should at least have a way to show some labels (e.g percentages) on the pie chart itself, no ?

spenserblack: I was worried that labels on the chart itself would take up too much space inside the chart. Once I fix the horizontal spacing, I'll test it out, but I'm worried it could result in smaller circles having too many "pixels" taken up by labels to be readable.

spenserblack: I could probably make it optional whether percentages/labels are rendered or not.

src/widgets/piechart.rs
```diff
@@ -0,0 +1,109 @@
+use std::cmp::min;
+use std::f64::consts::PI;
+
+use crate::buffer::Buffer;
+use crate::layout::Rect;
+use crate::style::{Color, Style};
+use crate::symbols::bar;
+use crate::widgets::{Block, Widget};
+
+const TAU: f64 = PI * 2.0;
```
Nytelife26, line 10: This is unnecessary as Τ is already defined under `std::f64::consts::TAU`.

spenserblack:
Since I haven't touched this PR in a long time I'm going to close it. If anyone wants to continue where I left off, I'll keep the branch open and a fork of my fork can be created.

PR #365, opened 2020-08-25
Implement highlighting of the selected row

dvoytik:
Introduce Table::highlight_row() switch to enable highlighting
of the selected row by applying the highlight style for all symbols
in the row (including spaces).

Signed-off-by: Dmitry Voytik <voytikd@gmail.com>

### DIFF
--- a/src/widgets/table.rs
+++ b/src/widgets/table.rs
@@ -101,6 +101,8 @@ pub struct Table<'a, H, R> {
     highlight_symbol: Option<&'a str>,
     /// Data to display in each row
     rows: R,
+    /// Use background highlight style to highlight the selected row
+    highlight_row: bool,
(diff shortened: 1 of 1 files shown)


dvoytik:
An example:
image

src/widgets/table.rs
```diff
@@ -188,6 +192,13 @@ where
         self
     }
 
+    /// Use background highlight style to highlight the selected row.
+    /// Background of the highlight style have be non-default for visible effect.
+    pub fn highlight_row(mut self, highligh: bool) -> Table<'a, H, R> {
```
Zooce, line 197: [super minor] missing a `t` in the parameter name `highligh`

fdehau:

PR #371, opened 2020-09-09
Add termchat to "apps using tui"

lemunozm:
Hi! 

I made [termchat](https://github.com/lemunozm/termchat) using `tui-rs`. I added it to the "apps using tui" list, but feel free to consider if it should be there because it seems simpler than the rest of the applications of the list.

Thank you for your effort in building `tui`!


fdehau:
There is a little conflict given that somebody else has added its app to the list. I'd be happy to add your project.

PR #377, opened 2020-09-19
Gauge uses Unicode Blocks

DashEightMate:
This pr allows the Gauge to display more detailed progress by using [Unicode block elements](https://en.wikipedia.org/wiki/Block_Elements). The changes don't have any bearing on outward-facing APIs or any other Gauge function. When there is a conflict between the label and the block character, the label character will display as normal, but with the colors from the ratio rounded down.


src/widgets/gauge.rs
```diff
@@ -129,6 +134,20 @@ impl<'a> Widget for Gauge<'a> {
     }
 }
 
+fn get_unicode_block<'a>(frac: f64) -> &'a str {
+    match (frac*8.0).round() as u16{ //get how many eighths the fraction is closest to
+        1 => "▏",
```
fdehau, line 139: Those graphemes are usually defined in [src/symbols.rs](https://github.com/fdehau/tui-rs/blob/master/src/symbols.rs) in order to potentially be used by other widgets.

src/widgets/gauge.rs
```diff
@@ -113,6 +113,11 @@ impl<'a> Widget for Gauge<'a> {
                 buf.get_mut(x, y).set_symbol(" ");
             }
 
+            //set unicode block
+            if self.ratio < 1.0{
+                buf.get_mut(end, y).set_symbol(get_unicode_block(width%1.0));
```
fdehau, line 118: I would add an option to use the old low resolution behavior as it is always wanted in environments where those unicode symbols are not well supported.

fdehau (changes requested):
It looks like a nice improvement. I've left some initial comments. I would like to see some tests too :).

DashEightMate:
Just made the gauge use the symbols in symbols.rs, and made an option to not use unicode. There's now tests for both with and without unicode, and the bottom gauge on the gauge example has unicode disabled as well.

src/widgets/gauge.rs
DashEightMate: Now references the symbols in symbols.rs as of the latest commit

src/widgets/gauge.rs
DashEightMate: Latest commit adds an option to revert to the old behavior, and I added a test for both new and old behavior

tests/widgets_gauge.rs
```diff
@@ -28,6 +28,46 @@ fn widgets_gauge_renders() {
             f.render_widget(gauge, chunks[1]);
         })
         .unwrap();
+    let expected = Buffer::with_lines(vec![
+        "                                        ",
+        "                                        ",
+        "  ┌Percentage────────────────────────┐  ",
+        "  │              ▋43%                │  ",
```
fdehau, line 35: Nice, could we also check the color behavior ? Especially when we are at 50% and the label may have different colors.

src/widgets/gauge.rs
```diff
@@ -129,6 +154,21 @@ impl<'a> Widget for Gauge<'a> {
...
+    match (frac * 8.0).round() as u16 {
+        //get how many eighths the fraction is closest to
+        1 => block::ONE_EIGHTH,
+        2 => block::ONE_QUARTER,
+        3 => block::THREE_EIGHTHS,
+        4 => block::HALF,
+        5 => block::FIVE_EIGHTHS,
+        6 => block::THREE_QUARTERS,
+        7 => block::SEVEN_EIGHTHS,
+        8 => "",
```
fdehau, line 167: shouldn't this be a full block ?

src/widgets/gauge.rs
```diff
@@ -81,6 +84,11 @@ impl<'a> Gauge<'a> {
         self.gauge_style = style;
         self
     }
+
+    pub fn use_unicode(mut self, unicode: bool) -> Gauge<'a> {
+        self.use_block = unicode;
```
fdehau, line 89: Any reason to have the property name be different from the name of the method setting its value ?

tests/widgets_gauge.rs
DashEightMate: I'm not sure how to get the color from the terminal/backend for an assert. Is there an example of this somewhere else?

src/widgets/gauge.rs
DashEightMate: There isn't a difference, I changed it over so it makes more sense with readability

src/widgets/gauge.rs
DashEightMate: Nope, just changed it over

tests/widgets_gauge.rs
fdehau: You can simply modify the cell of the expected `Buffer`. See [this](https://github.com/fdehau/tui-rs/blob/5a9b59866bbe93dad00309e4ffc449a21601972d/tests/widgets_list.rs#L33) for an example.

fdehau:
@DashEightMate Sorry for keeping you waiting. Beside the test on colors (I answered your comment about that) and the merge conflict, everything looks good. I would be more than happy to merge this if you are still up for it :).

DashEightMate:
@fdehau No worries! Just added color checks to the gauge as well as merged with main. Let me know if this needs anything else.

PR #392, opened 2020-09-29
Upgrade crossterm to v0.18

Byron:
Crossterm v0.18 reduces the amount of dependencies, among other improvements. I couldn't detect a breaking [change](https://github.com/crossterm-rs/crossterm/blob/master/CHANGELOG.md).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ unicode-segmentation = "1.2"
 unicode-width = "0.1"
 termion = { version = "1.5", optional = true }
 rustbox = { version = "0.11", optional = true }
-crossterm = { version = "0.17", optional = true }
+crossterm = { version = "0.18", optional = true }
 easycurses = { version = "0.12.2", optional = true }
 pancurses = { version = "0.16.1", optional = true, features = ["win32a"] }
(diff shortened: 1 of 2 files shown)


Byron:
Please note that easycurses also provides a new minor version, but I didn't want to conflate the PR as the only thing I tested was the crossterm backend.

fdehau:
Thanks ! I haven't found anything in the CHANGELOG suggesting a breaking change either. The demo looks good too. I'm happy to merge.

Regarding the easycurses backend I'm aware of the pending update but haven't found to investigate if it was safe to update. I'm also considering dropping support for this backend as I'm not sure if anybody is actually using it.

Byron:
Thanks a bunch!

> I'm also considering dropping support for this backend as I'm not sure if anybody is actually using it.

From all I can tell, this backend is the only one which [handles backgrounding](https://github.com/Byron/dua-cli/issues/65) and foregrounding properly. As far as I can tell, neither termion nor crossterm provide a way to handle the signals needed to do that.

Byron:
Since `crossterm` didn't document any breaking changes, would you mind publishing a patch release of `tui` as well to allow benefiting from `crossterm` 0.18?
Thank you

PR #398, opened 2020-10-15
feat(widgets/listitem): derive PartialEq

acheronfail:
While writing tests for my application I found it made things _a lot easier_ if I added `PartialEq` to `ListItem`. This way I can run assertions with an expected `Vec<ListItem>` against those that my app returns to verify everything is working correctly.

### DIFF
--- a/src/widgets/list.rs
+++ b/src/widgets/list.rs
@@ -36,7 +36,7 @@ impl ListState {
     }
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 pub struct ListItem<'a> {
     content: Text<'a>,
     style: Style,


acheronfail:
I worked around this by having my app return `Vec<Spans>`, and then translating that to `ListItem`s like:

```rs
self
    .nodes
    .iter()
    .flat_map(|node| {
        node.to_span_lines(&ctx) // Vec<Spans>
            .into_iter()
            .map(|spans| ListItem::new(spans))
            .collect::<Vec<ListItem>>()
    })
    .collect::<Vec<ListItem>>()
```

So if you don't want to derive `PartialEq` for `ListItem` then you can close this Pull Request. 🙂

fdehau:
It seems like a totally reasonable change, especially for a public type. Merged even though you have found a workaround :)

PR #399, opened 2020-10-15
fix: avoid gauge blinking

alocquet:
This PR fix a bug reproduced in gitui project (https://github.com/extrawurst/gitui/issues/328).

The gauge color blinks (see attached screencast): 
gitui-gauge-blinks

I think the problem comes with style cascading. I have replaced Gauge::default().style() with Gauge::default().gauge_style() but the gauge still blinks.

To fix this bug, I have reproduced code that is done for LineGauge :
gitui-gauge-ok


alocquet:
fix #400

Issue #404, opened 2020-10-31
Incorrect rendering of trailing space and NBSP in text reflow


pm100:
I want to show a text cursor as people are typing. So I set Spans that have "text so far", and " " with the trailing space underlined. If the user moves the cursor back then the current character is underlined. This is done in a Paragraph widget. But if Wrap is set on then the trailing space (and hence its underline is stripped out).

I read the code and saw that NBSP is treated specially, this changed the behavior but is still broken.

Here is a test app and its output

```
use std::io;
use std::io::Read;
use tui::backend::CrosstermBackend;
use tui::layout::{Constraint, Direction, Layout};
use tui::style::*;
use tui::text::*; //{Sp}
use tui::widgets::{Block, Borders, Paragraph, Widget, Wrap};
use tui::Terminal;
const NBSP: &str = "\u{00a0}";
fn main() -> Result<(), io::Error> {
    let stdout = io::stdout();
    let backend = CrosstermBackend::new(stdout);
    let mut terminal = Terminal::new(backend)?;
    terminal.draw(|f| {
        let chunks = Layout::default()
            .direction(Direction::Vertical)
            .margin(1)
            .constraints(
                [
                    Constraint::Percentage(50),
                    Constraint::Percentage(50),
                ]
                .as_ref(),
            )
            .split(f.size());

        let text = vec![
            Spans::from(vec![
                Span::raw("NBSP"),
                Span::styled(
                    NBSP.to_owned(),
                    Style::default().add_modifier(Modifier::UNDERLINED),
                ),
            ]),
            Spans::from(vec![
                Span::raw("Space"),
                Span::styled(
                    " ".to_owned(),
                    Style::default().add_modifier(Modifier::UNDERLINED),
                ),
            ]),
            Spans::from(vec![
                Span::raw("Underline"),
                Span::styled(
                    "_".to_owned(),
                    Style::default().add_modifier(Modifier::UNDERLINED),
                ),
            ]),
        ];
        let para = Paragraph::new(text.clone())
            .block(Block::default().title("Paragraph").borders(Borders::ALL))
            .style(Style::default().fg(Color::White).bg(Color::Black))
            .wrap(Wrap { trim: false });

        f.render_widget(para, chunks[0]);

        let para_no_wrap = Paragraph::new(text)
        .block(Block::default().title("Paragraph").borders(Borders::ALL))
        .style(Style::default().fg(Color::White).bg(Color::Black));

[...]

PR #405, opened 2020-10-31
Fix mishandling of NBSP with wrap turned on

pm100:
see issue #404

### DIFF
--- a/tests/widgets_paragraph.rs
+++ b/tests/widgets_paragraph.rs
@@ -2,7 +2,7 @@ use tui::{
     backend::TestBackend,
     buffer::Buffer,
     layout::Alignment,
-    text::{Spans, Text, Span},
+    text::{Span, Spans, Text},
     widgets::{Block, Borders, Paragraph, Wrap},
     Terminal,
 };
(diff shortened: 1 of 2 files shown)


fdehau:
Nice catch :). It seems we are missing a test on this case. Would you mind adding one ?

pm100:
added test

passing comment. Its really a pain that you have termion as default backend, working on tui on windows is always v fiddly

tests/widgets_paragraph.rs
```diff
@@ -141,6 +141,32 @@ fn widgets_paragraph_renders_mixed_width_graphemes() {
     terminal.backend().assert_buffer(&expected);
 }
 
+#[test]
+fn white_space_wrap(){
```
fdehau, line 145:
```suggestion
fn widgets_paragraph_can_wrap_with_a_trailing_nbsp() {
```
to be consistent with other tests and quickly launch all tests linked to a widget :)

fdehau:
> added test

Thanks :+1:  Rustfmt is not happy though.

> Its really a pain that you have termion as default backend, working on tui on windows is always v fiddly

I understand and `crossterm` will probably become the default backend at some point but I'm waiting for it to stabilize a bit.

Issue #90, opened 2018-11-17
Allow Lists to wrap


svenstaro:
I would like to present stuff in a list but I'd also like stuff in it to be able to wrap long strings.

fdehau:
Somewhat linked to #81. A rework of the `List` and `SelectableList` widgets should take into account both feature requests.

PR #413, opened 2020-11-14
feat(text): add a way to wrap text without Paragraph

fdehau:


VanillaBrooks:
Looks like this would solve #90 which is a problem I'm currently facing. Whats left to complete here @fdehau? Perhaps I can take a stab at it

Wuelle:
> 
> 
> Looks like this would solve #90 which is a problem I'm currently facing. Whats left to complete here @fdehau? Perhaps I can take a stab at it

+1

avdb13:
Bump?

Issue #395, opened 2020-10-07
Block with height=1 doesn't display content


famz:
Is there a reason why we reject 1 row or 1 col `Block`? I'm thinking about such a change (should be conditional on border style == NONE), just so that I can use Block to display a custom line.

```
diff --git a/src/widgets/block.rs b/src/widgets/block.rs
index e168698..d207766 100644
--- a/src/widgets/block.rs
+++ b/src/widgets/block.rs
@@ -137,7 +137,7 @@ impl<'a> Widget for Block<'a> {
     fn render(self, area: Rect, buf: &mut Buffer) {
         buf.set_style(area, self.style);

-        if area.width < 2 || area.height < 2 {
+        if area.width < 1 || area.height < 1 {
             return;
         }

```

fdehau:
You are right, there is no valid reason to not render `Block` of width 1 or height 1. Fixed in #415.

PR #415, opened 2020-11-14
fix(widgets/block): allow Block to render on small areas

fdehau:
Allow `Block` to render on areas where the height or the width is 1.


fdehau:
Fix #395

PR #425, opened 2020-12-05
Add comments to "list" example and fix list direction

Nukesor:
Hi :)

First of all, thanks for providing this crate! I'm really looking forward to build stuff with it :) 

This PR only adds some comments to the `list` example. Hopefully, this will make it easier for new people to understand those.
The example code is sometimes a little overwhelming, when one only wants to get a brief overview on how the library works, without deepdiving into the docs.

On top of that, this fixes the flow direction of the event list (it was previously going from 26 to 1)

### DIFF
--- a/examples/list.rs
+++ b/examples/list.rs
@@ -16,10 +16,9 @@ use tui::{
     Terminal,
 };
 
-/// This struct holds the current state of the app.
-/// The `items` field fulfills all prerequisites for Tui's `StatefulWidget` trait.
-/// This achieves that Tui can, for instance, understand the currently selected item and can
-/// automatically highlight it.
+/// This struct holds the current state of the app. In particular, it has the `items` field which is a wrapper
+/// around `ListState`. Keeping track of the items state let us render the associated widget with its state
+/// and have access to features such as natural scrolling.
 ///
 /// Check the event handling at the bottom to see how to change the state on incoming events.
 /// Check the drawing logic for items on how to specify the highlighting style for selected items.
@@ -90,7 +89,7 @@ impl<'a> App<'a> {
     }
 
     /// Rotate through the event list.
-    /// This only exists go simulate some kind of "progress"
+    /// This only exists to simulate some kind of "progress"
     fn advance(&mut self) {
         let event = self.events.remove(0);
         self.events.push(event);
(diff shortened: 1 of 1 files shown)


examples/list.rs
```diff
@@ -131,19 +145,26 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+                .map(|&(event, level)| {
+                    // Colorcode the level depending on it's type
                     let s = match level {
                         "CRITICAL" => Style::default().fg(Color::Red),
                         "ERROR" => Style::default().fg(Color::Magenta),
                         "WARNING" => Style::default().fg(Color::Yellow),
                         "INFO" => Style::default().fg(Color::Blue),
                         _ => Style::default(),
                     };
+                    // Add a exapmle datetime and apply proper spacing between them
```
fdehau, line 167:
```suggestion
                    // Add a example datetime and apply proper spacing between them
```

examples/list.rs
```diff
@@ -131,19 +145,26 @@ fn main() -> Result<(), Box<dyn Error>> {
...
 
+            // Let's do the same for the events.
+            // The event list doesn't have any state and only displays the current state of the list.
             let events: Vec<ListItem> = app
                 .events
                 .iter()
-                .map(|&(evt, level)| {
+                .rev()
+                .map(|&(event, level)| {
+                    // Colorcode the level depending on it's type
```
fdehau, line 159:
```suggestion
                    // Colorcode the level depending on its type
```

examples/list.rs
```diff
@@ -16,6 +16,13 @@ use tui::{
     Terminal,
 };
 
+/// This struct holds the current state of the app.
+/// The `items` field fulfills all prerequisites for Tui's `StatefulWidget` trait.
+/// This achieves that Tui can, for instance, understand the currently selected item and can
+/// automatically highlight it.
```
fdehau, line 22:
```suggestion
/// This struct holds the current state of the app. In particular, it has the `items` field which is a wrapper
/// around `ListState`. Keeping track of the items state let us render the associated widget with its state
/// and have access to features such as natural scrolling. 
```

fdehau (commented):
Thanks for taking the time to do this. I've left some suggestions. Let me know what you think.

Nukesor:
Ah perfect.
Seems like I got the implementation details of `StatefulList` and the handling of a widget's state wrong. Thanks for clarifying this :)

Your suggested changes have been applied and a few additional typos have been fixed.

PR #426, opened 2020-12-06
Draw table header and border even if rows are empty

vchekan:
If table rows are empty, it does not render at all.
I would expect header and empty border to be rendered.

### DIFF
--- a/src/widgets/table.rs
+++ b/src/widgets/table.rs
@@ -402,9 +402,6 @@ impl<'a> StatefulWidget for Table<'a> {
             return;
         }
         buf.set_style(area, self.style);
-        if self.rows.is_empty() {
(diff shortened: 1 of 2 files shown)


fdehau (commented):
Good catch ! Could you add a test as well so we don't forgot this case in the future ?

vchekan:
Done. 
Should I squash those commits? Or you'll do it when merging?

P.S. windows build is failing because can not find `make`?

fdehau (approved):

fdehau:
@vchekan LGTM. You are right I broke the windows build in #428 and fixed in #430, a rebase should fix it. And yes you can either squash your commits yourself or I'll do it when merging.

PR #429, opened 2020-12-07
Make examples available for all platforms

sagiegurari:
Really cool project. I use it internally :)
I saw that the windows examples task i not implemented.
So i played with it a bit to make it work.
I couldn't test it on windows (need your help if possible) but on linux it works and its same code now (no need for 2 different tasks).
Tell me what you think?

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -93,19 +93,13 @@ args = [
 ]
 
 [tasks.run-examples]
-linux_alias = "run-examples-unix"
-mac_alias = "run-examples-unix"
-windows_alias = "run-examples-windows"
-
-[tasks.run-examples-unix]
 script = '''
-#!/usr/bin/env bash
-cargo build --examples --features ${TUI_FEATURES} --release
-for file in examples/*.rs; do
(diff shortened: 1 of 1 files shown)


fdehau:
Thanks for looking into this and for `cargo-make` :). I would be more than happy to merge this but I see two issues:
1. While it works for you, the examples are not responsive on Linux for me: pressing 'q' does not exit the example. I don't know if this can be related to how exec work in duckscript.
2. Currently most examples do not compile on Windows because we use `termion` as a default backend. I plan to move to `crossterm` for the default backend at some point to improve this. In the meantime, it would be great if `run-examples` would only run the `crossterm_demo` example.

sagiegurari:
@fdehau thanks for checking. I'll try to see what i can do and keep it working for all platforms in parallel :)

fdehau:
Great it works perfectly now. I like that your solution is also a proof of `cargo-make` flexibility. Thanks again for taking the time to look into this.

sagiegurari:
thanks. feel free to ping me if you need any help/support related to the build script.

PR #434, opened 2020-12-20
Add readme to Cargo.toml so that crates.io shows the README.md instead of an empty page

phip1611:
Hi. Cargo.toml is missing the "readme" field. This results in an empty page (one paragraph only) on crates.io https://crates.io/crates/tui

If you add this field/key the README.md will be visible on crates.io, like here: https://crates.io/crates/bat

This would attract more users if they immidately can see screenshots etc.

Happy holidays/merry christmas!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,8 @@ description = """
 A library to build rich terminal user interfaces or dashboards
 """
 documentation = "https://docs.rs/tui/0.10.0/tui/"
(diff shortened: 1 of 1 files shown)


fdehau:
I'm a bit surprised this is required given that the documentation [mentions](https://doc.rust-lang.org/cargo/reference/manifest.html#the-readme-field) it should default to `README.md` if the file exists. Anyway, I've added it as part of #438, so I'm closing this PR but thanks for reporting it :).

PR #453, opened 2021-02-14
Snake example

Valerioageno:
Hello guys. I've implemented the "fill" option in the rectangle canvas struct and create the "snake game" example.  
I hope it's going well!

Have a nice week!
Best regards
Valerio

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,11 @@ path = "examples/canvas.rs"
 required-features = ["termion"]
 
 [[example]]
+name = "snake"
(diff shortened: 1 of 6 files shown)


fdehau:
Hey ! Thank for your interest in `tui`. This is a nice looking example but I'm not looking into adding more examples at the moment given that they are already quite a lot of them and the README includes a list of apps using `tui` which may showcase advanced usage.
Also the change to the `Rectangle` canvas should probably get its own PR with a quick description on why the change is needed, documentation and tests :).

PR #454, opened 2021-02-23
Render widgets without consuming them

sivizius:
Adds two methods (`render` and `render_stateful`) to render widgets without consuming them.
Signature of Widgets:render are changed therefore from `(self, …)` to `(&mut self, …)`, which breaks custom widgets a wee bit.
This allows to construct widgets in before and have them in an interface-object, modify only field directly or via additional methods on the widgets if necessary and render them without allocate memory every time. Yet the old way with `render_widget` and `render_stateful_widget` is still possible, but I sometimes prefer to have some kind of `init()`, `update()` and `draw()`.


fdehau:
First of all thanks for your interest in this crate. I'm a bit surprised by the scale of the PR without prior discussion as well as the inclusion of what look like unrelated changes. This is basically shifting the paradigm of this crate from an immediate mode (which was intended) to a more retained approach which already exist in other crates (see `cursive` for example). Therefore, I won't be able to review nor accept such PR at the moment. I'm still encouraging you to open issues if you want to discuss API designs.

PR #459, opened 2021-03-22
Fix out of bounds table offset panic

BenHall-7:
So I have a TUI project I'm working on where I've supplied a method to filter out data from the Table widget, using a regex match. Whenever I do this I want to preserve the previous table state, so that the table can still scroll up or down relative to its previous position. For example, if just 1 entry were removed it would be preferable to preserve the offset instead of reset it to 0 and scroll down to the currently selected element. The issue this led to was an out-of-bounds index whenever the offset was greater than the length of the new data. The fix takes only 1 line: clamp the offset to the last index in the new list.

### DIFF
--- a/src/widgets/table.rs
+++ b/src/widgets/table.rs
@@ -334,7 +334,7 @@ impl<'a> Table<'a> {
         offset: usize,
         max_height: u16,
     ) -> (usize, usize) {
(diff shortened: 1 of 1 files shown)


fdehau:
Thanks a lot for the detailed report. I've fixed this in #516  to add tests and fix it in `List` which had the same issue.

PR #461, opened 2021-03-26
Actually clear buffer in TestBackend::clear

scvalex:
The `TestBackend::clear` function didn't actually do anything before.  This feature changes it to clear the buffer.

I ran into this because I'm testing my full application (instead of just specific widgets), and switching between screens (which have `terminal.clear()` calls in between) was leaving junk on the screen.


fdehau:
Good catch. Thanks for looking into it 🙇‍♂️ .

Issue #450, opened 2021-02-01
Question: Set the title to the top right or middle of a block.


KallenW:
I didn't find the relevant API in the documentation. Maybe I should implement a custom widget ?

fdehau:
This is not supported at the moment but should be fairly easy to implement, e.g adding a `title_alignment` to `Block` and change the logic [here](https://github.com/fdehau/tui-rs/blob/4e76bfa2ca8eb51719d611bb8d3d4094ab8ba398/src/widgets/block.rs#L195) to handle this new property.

I might have the time to address it in a future release but in the meantime I'll leave it as a good first issue if anyone want to pick it up.

olitus:
Hello, I am kinda newbie in Rust and would like to implement this. 
This would be my first contribution to Open Source project, so I am very excited.
Thank you for pointing to the correct parts of code.

olitus:
Here is my pull request https://github.com/fdehau/tui-rs/pull/462

oreksu:
Should we close this issue?

PR #462, opened 2021-03-28
Added feature to align title of the block

oreksu:
This is PR for https://github.com/fdehau/tui-rs/issues/450

- Added title_alignment field to block struct
- Added title_alignment() method to set title_alignment of the block
- Added comments and title alignment to the example/block.rs
- Added tests for title alignment in tests/widgets_block.rs
- Added tests in srs/block.rs for inner

For all above, used layout::Alignment enum as the type for title_alignment.

### DIFF
--- a/src/widgets/block.rs
+++ b/src/widgets/block.rs
@@ -1,6 +1,6 @@
 use crate::{
     buffer::Buffer,
-    layout::{Rect, Alignment},
+    layout::{Alignment, Rect},
     style::Style,
     symbols::line,
     text::{Span, Spans},
@@ -203,7 +203,6 @@ impl<'a> Widget for Block<'a> {
 
         // Title
         if let Some(title) = self.title {
-
             let left_border_dx = if self.borders.intersects(Borders::LEFT) {
                 1
             } else {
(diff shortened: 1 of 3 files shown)


src/widgets/block.rs
```diff
@@ -507,5 +532,39 @@ mod tests {
                 height: 0,
             },
         );
+        assert_eq!(
```
fdehau, line 535: The `title_alignment` is not taken into account in `Block::inner`. Therefore, is there a reason to add those test cases here ?

tests/widgets_block.rs
```diff
@@ -211,3 +211,471 @@ fn widgets_block_renders_on_small_areas() {
         Buffer::with_lines(vec!["┌Test─"]),
     );
 }
+
+#[test]
+fn widgets_block_renders_title_top_left_all_borders() {
```
fdehau, line 216: Thanks for including tests. However, a lot of them share the same structure. Could we use a more data-driven approach here ? For example, in [widgets_paragraph_can_wrap_its_content](https://github.com/fdehau/tui-rs/blob/1028d39db0946736b3c11a7c641f425dbf269030/tests/widgets_paragraph.rs#L16), the common structure is extracted to a closure which is then called with all test inputs. Additionally, the `TestBackend` could maybe a bit smaller to have more concise buffer assertions.

fdehau (changes requested):
Sorry for keeping you waiting for a review. This is a very nice pull request. I have some suggestions to make it even better. Let me know what you think :).

src/widgets/block.rs
oreksu: I added them to show that `title_alignment` does not effect `Block::inner`. I am not sure, if that is necessary.

src/widgets/block.rs
```diff
@@ -192,19 +201,35 @@ impl<'a> Widget for Block<'a> {
...
-            buf.set_spans(area.left() + lx, area.top(), &title, width);
+
+            let title_area_width = area
+                .width
+                .saturating_sub(left_border_dx)
+                .saturating_sub(right_border_dx);
+
+            let title_dx = match self.title_alignment {
+                Alignment::Left => 0,
+                Alignment::Center => (title_area_width - title.width() as u16) / 2,
```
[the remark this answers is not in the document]
oreksu: I will update to use `saturating_sub` and add a test for this case. Thank you for pointing this out.

tests/widgets_block.rs
oreksu: Ohhhh, that is so neat! Thank you for showing the closure example, I will try to do something similar. 
I will make `TestBackend` smaller.

oreksu (commented):
Thank you for all of the suggestions :)

oreksu:
Committed updates based on the suggestions and fixed behavior of the center alignment when used without left border.

fdehau (approved):

PR #467, opened 2021-04-12
Give the crate::widgets::reflow module public visibility

jjpe:
I'm writing a new widget for `tui-rs` ecosystem that I eventually plan to open-source.
But the new widget needs access to either the `widgets::reflow` library, or copy the code.
Reuse seems like the better option to me.

### DIFF
--- a/src/widgets/mod.rs
+++ b/src/widgets/mod.rs
@@ -23,7 +23,7 @@ mod clear;
 mod gauge;
 mod list;
 mod paragraph;
-mod reflow;
+pub mod reflow;
 mod sparkline;
 mod table;
 mod tabs;


fdehau:
Hey 👋 . This change is quite important in the sense it increases significantly the API surface of the crate and thus the maintenance cost. Could you provide more details on your needs ?

I think this is also somewhat linked to the missing ability to have the size of a text after being wrapped in `Paragraph`. For example, a more flexible `Text` object with wrapping and alignment capabilities might help you. I'll try to open an issue with a summary of possible solutions.

jjpe:
Hi there :smile:

Here's the use case:
I've written a `TextEditor` widget for `tui-rs`, which I'd like to open source and distribute via crates.io.
In order to write it, I used the ` tui::widgets::Paragraph` type as an example, in particular `impl StatefulWidget for Paragraph`.

The thing is, `impl StatefulWidget for Paragraph` makes the exact same use of these 3 types as my `TextEditor` widget does:

```rust
tui::widgets::reflow::{LineComposer, LineTruncator, WordWrapper}
```

PR #471, opened 2021-04-17
feat(Text): Add a From<Cow<str>> impl for Text

eminence:
This is useful when you have a list of `Cow<str>` and you want to use it in a Row.

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -398,6 +398,12 @@ impl<'a> From<&'a str> for Text<'a> {
     }
 }
 
+impl<'a> From<Cow<'a, str>> for Text<'a> {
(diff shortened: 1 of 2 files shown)


fdehau:
Hey 👋 . Could you add a test (maybe a doc test) to make sure this case is covered going forward ?

eminence:
Sure!  I pushed a new commit that add two examples.  One is in the docs for `Cell`, but this was kinda boring so I also added another example in the docs for `Row` (which happen to mirror my own personal use case):

```rust
Row::new(vec![
    Cow::Borrowed("hello"),
    Cow::Owned("world".to_uppercase()),
]);
```

PR #497, opened 2021-06-17
state: adds offset access and manipulation

### LINKED ISSUE #495: manipulating ListState/TableState offset
## Problem

The `offset` of `ListState` and `TableState` can neither be accessed nor manipulated. 

## Solution

Add publicly visible `fn offset` and `fn offset_mut` to both `ListState` and `TableState`. 

## Alternatives

None.

## Additional context

Offset manipulation is needed to implement controls to center the view around the currently selected list item or row, e.g. like Emacs' `recenter-top-bottom` via `C-l` shortcut.

wookietreiber:
## Description

Adds offset access and manipulation. Fixes #495. 

## Testing guidelines

I have a few structs that keep the table state. I use this `trait ScrollableTable` for each of these to attach actions like scroll up/down, etc. I've added this `fn recenter` to test offset manipulation. `ListState` should be pretty much analogously. 

Note that I haven't added tests per se, because I think that these kinds of accessor/mutator functions are quite obvious. 

```rust
use tui::layout::Rect;
use tui::widgets::TableState;

pub trait ScrollableTable {
    fn size(&self) -> Rect;
    fn state(&self) -> &TableState;
    fn state_mut(&mut self) -> &mut TableState;

    fn recenter(&mut self) {
        if let Some(selected) = self.state().selected() {
            let height = self.size().height as usize;

            let offset = self.state_mut().offset_mut();
            *offset = selected.saturating_sub(height / 2);
        }
    }
}
```

## Checklist

### DIFF
--- a/src/widgets/list.rs
+++ b/src/widgets/list.rs
@@ -14,6 +14,16 @@ pub struct ListState {
 }
 
 impl ListState {
+    /// Returns the offset.
+    pub fn offset(&self) -> usize {
+        self.offset
+    }
+
+    /// Mutably returns the offset.
+    pub fn offset_mut(&mut self) -> &mut usize {
+        &mut self.offset
+    }
+
     pub fn selected(&self) -> Option<usize> {
         self.selected
     }
--- a/src/widgets/table.rs
+++ b/src/widgets/table.rs
@@ -342,6 +342,16 @@ pub struct TableState {
 }
 
 impl TableState {
+    /// Returns the offset.
+    pub fn offset(&self) -> usize {
+        self.offset
+    }
+
+    /// Mutably returns the offset.
+    pub fn offset_mut(&mut self) -> &mut usize {
+        &mut self.offset
+    }
+
     pub fn selected(&self) -> Option<usize> {
         self.selected
     }


src/widgets/list.rs
```diff
@@ -23,6 +23,16 @@ impl Default for ListState {
 }
 
 impl ListState {
+    /// Returns the offset.
```
Ranguna, line 26: Do we need these comments ?

Asking because the rest of the code doesn't seem to use comments like these.

wookietreiber: Even if there aren't many doc comments as of now, why not start?

Ranguna: I personally prefer comments only when strictly needed, the code you wrote looks self explanatory to me, so we don't need to maintain code + comments.

But I'll leave this to the maintainer to review 👌

wookietreiber:
cargo make does not seem to be compatible with rustc 1.56.1: 

```
error: failed to compile `cargo-make v0.35.8`, intermediate artifacts can be found at `/tmp/cargo-install52QPLl`

Caused by:
  package `os_str_bytes v6.2.0` cannot be built because it requires rustc 1.57.0 or newer, while the currently active rustc version is 1.56.1
Error: Process completed with exit code 101.
```

PR #502, opened 2021-06-23
Add joshuto to the "apps using tui" list

kamiyaa:
`joshuto` is a ranger-like terminal file manager written in Rust


fdehau:
Done in #544 :)

kamiyaa:
Thanks!

PR #505, opened 2021-07-05
Make bitflags smaller

cbarrete:
## Description

This reduces the size of bitflags so that they are less wastful. They can be grown again later if required.

## Testing guidelines

`cargo test` still passes, the behavior is unchanged.

## Checklist

* [X] I have documented all new additions.


fdehau (changes requested):
I don't agree with the `Modifier` change but the `Borders` one looks ok.

Issue #470, opened 2021-04-16
Table widget panics when table doesnt fit screen


deepu105:
**Describe the bug**
The table widget is panicking when it doesn't for the screen

Here is the stack trace thrown at https://github.com/kdash-rs/kdash/blob/main/src/ui/overview.rs#L262
```
thread '<unnamed>' panicked at 'Trying to access position outside the buffer: x=189, y=16, area=Rect { x: 0, y: 0, width: 189, height: 40 }', /home/deepu/.cargo/registry/src/github.com-1ecc6299db9ec823/tui-0.14.0/src/buffer.rs:217:9
   0: kdash::panic_hook
             at src/main.rs:57:46
   1: kdash::main::{{closure}}::{{closure}}
             at src/main.rs:76:5
   2: std::panicking::rust_panic_with_hook
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/std/src/panicking.rs:581:17
   3: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/std/src/panicking.rs:484:9
   4: std::sys_common::backtrace::__rust_end_short_backtrace
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/std/src/sys_common/backtrace.rs:153:18
   5: rust_begin_unwind
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/std/src/panicking.rs:483:5
   6: std::panicking::begin_panic_fmt
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/std/src/panicking.rs:437:5
   7: tui::buffer::Buffer::index_of
             at /home/deepu/.cargo/registry/src/github.com-1ecc6299db9ec823/tui-0.14.0/src/buffer.rs:217:9
   8: tui::buffer::Buffer::get_mut
             at /home/deepu/.cargo/registry/src/github.com-1ecc6299db9ec823/tui-0.14.0/src/buffer.rs:184:17
   9: tui::buffer::Buffer::set_style
             at /home/deepu/.cargo/registry/src/github.com-1ecc6299db9ec823/tui-0.14.0/src/buffer.rs:358:17
  10: tui::widgets::table::render_cell
[...]

PR #508, opened 2021-07-21
Fix for table panic

andschneider:
## Description

This is an initial fix for issue #470.

After walking through one of the cases that caused the panic in a debugger, I found that it was due to a miscalculation (it seems like a rounding issue) in the column widths from the `cassowary` solver. I don't fully understand the parameters for this solver and as such I didn't feel comfortable trying to implement a fix into the `get_column_widths` function. Changing the solver also seems to have potential knock-on effects in how all tables are rendered, which sounds like a disruptive change.

So, I tried to make a minimal fix by simply checking for overflow in the `set_style` function and clamping the value if there would have been an overflow. 

I'm not exactly sure if this is a completely robust solution, but it seems to work for the two test cases we have so far.

@fdehau would love your thoughts on how robust of a fix this is!

## Testing guidelines

I added the two example tables that were causing a panic from the issue in `widgets_table_columns_dont_panic` test function. All this test does is render the tables and ensures a panic doesn't occur. It doesn't verify the output of the table.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## To be released
 
+### Fixes
+
+* Some combinations of `Table` widths and `Rect` width would cause a panic due to accessing area
+outside the buffer - (#470)
+
 ## v0.15.0 - 2021-05-02
 
 ### Features
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tui"
-version = "0.15.0"
+version = "0.15.1"
 authors = ["Florian Dehau <work@fdehau.com>"]
 description = """
 A library to build rich terminal user interfaces or dashboards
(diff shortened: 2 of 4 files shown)


andschneider:
So after a little more testing:

```rust
for i in 10..1000 {
    test_case(&mut state, table2.clone(), i);
}
```

I found that with table widths below 100, `table2` from the tests still panics sometimes. Not sure why, but this fix is definitely not completely robust. It does work above 100 though, and `table1` still does not panic.

deepu105:
I'll test this branch as soon as I find some time

deepu105 (approved):
This seems to work, atleast for the scenarios I tested. but regrdless its better than the current behaviour

deepu105:
@fdehau WDYT? it would be nice to have this in a patch release

fdehau:
Thanks for looking into it and sorry for the late response. I'm not very fond of the proposed fix because it can potentially hide a lot of bugs which will be harder to track and fix. Thanks to your provided regression test, I was able to identify an issue in the column widths computation. I've open [this PR](https://github.com/fdehau/tui-rs/pull/514) which reuses existing code and should fix the issue.

andschneider:
Thanks @fdehau! Glad there was a better solution :100:

Issue #526, opened 2021-08-18
Frame::render_widget() can't take Widget trait objects


SergeyKasmy:
## Problem

I'm trying to make a function that takes a closure with a signature similar to this one

```
Fn(Rect) -> Vec<(Box<dyn Widget>, Rect)>
```

and allows to display dynamic content. Currently there's no way to call render_widget() with a `dyn Widget`. There was a similar issue (#178) but my use case would be different. 

### Here's what I want to achieve in more detail:

I have a list of "Players" on the left with information about them on the right. Example:

![image_2021-08-01_23-37-36](https://user-images.githubusercontent.com/37976824/129832738-082aa97c-a125-48dd-8fa5-dbbe2483a7be.png)

I'd like to be able to show not only players but also different objects that would require to have different info shown on the right pane. I was thinking of a function like

```
pub fn draw_editor<F>(
	&self,
	list_title: Option<&str>,
	list_items: &[&str],
	details: Option<F>,
) -> Result<EditorAction>
where
	F: Fn(Rect) -> Vec<(Box<dyn Widget>, Rect)>,
```

that considers the type of the currently selected object/entities in the closure and display the appropriate info. Sadly I can't find a way to do a generic "editor" screen without using different functions for different types of objects displayed and duplicating a lot of code.

Is there anything I can do to avoid just copy-pasting everything 3 (or even more in the future) times?

SleeplessSloth:
Internet lagged out a bit, this posted too early, sorry! I'll edit this one to say what I wanted to say.

emlaufer:
Not sure if you are still having this issue, but why not just use enums for this? Each object type can just be a different variant. Then you can just write a render function which accepts the Frame as a parameter, match on which variant self is, then render based on the type of object. For example:

```
enum Object {
    Person { name: String, ... },
    OtherThing { name: String, ...},
}
impl Object {
    fn render<B>(&self, frame: &mut Frame<B>, area: Rect) 
    where B: Backend {
        match self {
            Object::Person { ... } => { /* render the person details, e.g. stats, skills, etc */ }
            Object::OtherThing { ... } => { /* render other things */ }
        }
    }
}
```

PR #528, opened 2021-08-18
Add a render method for the Widget trait that takes by reference instead

SergeyKasmy:
## Description
This PR adds `Terminal::render_widget_ref()` and `(Stateful)Widget::render_ref()` functions that take Widgets by reference instead of taking ownership. This allows to call render on a Trait object with unknown size with is useful where the exact object implementing Widget is unknown or depends on some kind of condition. I decided not to change the `render_widget() `and `render()` signatures as to keep backwards compatibility but sadly all custom widgets are still broken because they are now required to implement `render_ref()` instead. This is just a proof of concept PR and I would appreciate any help on getting it up to the standards since this is my first time contributing to a big rust project. I made this to fix issue #526 I posted yesterday and I'm already using this PR in a project of mine. 

## Checklist

### DIFF
--- a/examples/custom_widget.rs
+++ b/examples/custom_widget.rs
@@ -19,7 +19,7 @@ impl<'a> Default for Label<'a> {
 }
 
 impl<'a> Widget for Label<'a> {
-    fn render(self, area: Rect, buf: &mut Buffer) {
+    fn render_ref(&mut self, area: Rect, buf: &mut Buffer) {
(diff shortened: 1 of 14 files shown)


harrypunk:
Super helpful feature. I'm looking for a way of rendering a dynamic widget.

kroonhorstdino:
Would definitely like to see this feature merged! Would make my life so much easier, right now you have to work around so much limitations in that regard.

PR #551, opened 2021-10-30
Feature: Block padding

Igormandello:
## Description
If we want to render a widget inside a block with a certain distance from its borders, we need to create another `Layout` element based on the outer `Block`, add a margin and render the Widget into it. Adding a padding property on the block element skips the creation of this second Layout.

This property works especially when rendering texts, as we can just create a block with padding and use it as the text wrapper:

```rust
let block = Block::default().borders(Borders::ALL).padding(4);
let paragraph = Paragraph::new("example paragraph").block(block);
f.render_widget(paragraph, area);
```

Rendering another widget should be easy too, using the `.inner` method:

```rust
let block = Block::default().borders(Borders::ALL).padding(4);
let inner_block = Block::default().borders(Borders::ALL);
let inner_area = block.inner(area);

f.render_widget(block, area);
f.render_widget(inner_block, inner_area);
```

## Testing guidelines
The `block.rs` example contains two cases of use: one with an inner text and another with an inner block:
image

I've also changed the paragraph tests to include a padding of value 1 to the rendering tests.

## Checklist

* [X] I have documented all new additions.

### DIFF
--- a/examples/block.rs
+++ b/examples/block.rs
@@ -9,7 +9,7 @@ use tui::{
     layout::{Alignment, Constraint, Direction, Layout},
     style::{Color, Modifier, Style},
     text::Span,
-    widgets::{Block, BorderType, Borders},
+    widgets::{Block, BorderType, Borders, Paragraph},
     Terminal,
 };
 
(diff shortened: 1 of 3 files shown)


alpha-tango-kilo:
This would be awesome! I hope this gets merged 🙌

Issue #567, opened 2021-12-05
Y-Axis position in the Chart widget


theogilbert:
## Problem

This issue has to do with the layout of the `Chart` widget.
The position of the Y axis is offset by the maximum width of the labels on the left of the axis.

Currently, the first label of the X axis is considered as one of these labels.

If the labels of the Y axis are not very wide, but the first label of the X axis is, this will needlessly reduce the size of the chart, and add a lot of empty space to the UI.

## Solution

A simple solution, but one which could affect many users, would be to stop considering this X axis label to be on the left of the Y axis.

This would be done by updating the following methods:
- [`Chart::max_width_of_labels_left_of_y_axis()`](https://github.com/fdehau/tui-rs/blob/master/src/widgets/chart.rs#L356)
- [`Chart::render_x_labels()`](https://github.com/fdehau/tui-rs/blob/master/src/widgets/chart.rs#L372)

## Alternatives

In order to allow users to customize this behavior, and to not break their UI, we could offer them a way to customize this behavior.

Here are a couple of ideas on how this positioning could be customized by users:

1. The `Axis` structure could have a new method to define the alignment of its labels (left, center, right).
	- For the Y axis, it would be useful to align labels of different sizes.
	- For the X axis, it would define the alignment of the **first label** compared to the Y-axis.
	  If the alignment is **left** (*default*), the behavior would stay as it is today, **affecting** the position of the Y axis.
          If the alignment is **right**, the first label would be drawn after the Y axis, and would **not affect** its position.
          If the alignment is **center**, the first label would be centered on the position of the Y axis. The Y axis would only be offset by half the width of the label.
2. The `Chart` structure could have a new method to define the positioning of the first label of the X axis.
	For example: `Chart::x_label_alignment(alignment: Alignment)`

## Additional context

Current positioning of the first X axis label:
![image](https://user-images.githubusercontent.com/16946214/144727983-c72d38bd-eed3-4cbf-b9d2-77a455f66c20.png)

Wanted positioning:
![image](https://user-images.githubusercontent.com/16946214/144727994-ef68bd78-32de-40ba-a4da-e6a7a142c0ac.png)

PR #568, opened 2021-12-08
Chart axis alignment

theogilbert:
## Description

This PR allows users to customize the alignment of the chart widget's axis' labels.
It implements a solution of the issue #567.

The alignment can be customized through the new method `Axis::labels_alignment(alignment: Alignment)`:

```rust
let axis = Axis::default()
    .labels([Span::from("0"), Span::from("100")])
    .labels_alignment(Alignment::Right);
```

By default, the alignment of an `Axis` is `Alignment::Left`.
This produces an output similar to the one currently produced.

Defining the alignment of the Y axis will affect the position of all Y-axis labels.
Defining the alignment of the X axis will only affect the position of the first label, relative to the Y-axis.

A side-effect of the PR is that X-axis labels are now closer to their correct position.
Currently on the  branch `master`:
image
With this PR:
image

## Testing guidelines

New scenarios have been added in `tests/widgets_chart.rs` to test the output produced by different alignments.

### X-Axis label alignments:
`Alignment::Left` (*default*)
image
`Alignment::Center`
image
`Alignment::Right`
image

### Y-Axis label alignments:
`Alignment::Left` (*default*)
image
`Alignment::Center`
image
`Alignment::Right`
image

## Checklist

### DIFF
--- a/src/widgets/chart.rs
+++ b/src/widgets/chart.rs
@@ -412,36 +412,34 @@ impl<'a> Chart<'a> {
 
         let width_between_ticks = graph_area.width / labels_len;
 
-        for (i, label) in labels.iter().enumerate() {
-            if i == 0 {
-                let label_area = self.first_x_label_area(
-                    y,
-                    label.width() as u16,
-                    width_between_ticks,
-                    chart_area,
-                    graph_area,
-                );
+        let label_area = self.first_x_label_area(
+            y,
+            labels.first().unwrap().width() as u16,
+            width_between_ticks,
+            chart_area,
+            graph_area,
+        );
 
-                let alignment = match self.x_axis.labels_alignment {
-                    Alignment::Left => Alignment::Right,
-                    Alignment::Center => Alignment::Center,
-                    Alignment::Right => Alignment::Left,
-                };
+        let label_alignment = match self.x_axis.labels_alignment {
+            Alignment::Left => Alignment::Right,
+            Alignment::Center => Alignment::Center,
+            Alignment::Right => Alignment::Left,
+        };
 
-                Self::render_label(buf, label, label_area, alignment);
-            } else if i == labels.len() - 1 {
-                let x = graph_area.right() - width_between_ticks;
-                let label_area = Rect::new(x, y, width_between_ticks, 1);
-                // The last label should be aligned Right to be at the edge of the graph area
-                Self::render_label(buf, label, label_area, Alignment::Right);
-            } else {
-                // We do a +1 (and width-1 below) to leave at least one space before each intermediate labels
-                let x = graph_area.left() + i as u16 * width_between_ticks + 1;
-                let label_area = Rect::new(x, y, width_between_ticks.saturating_sub(1), 1);
+        Self::render_label(buf, labels.first().unwrap(), label_area, label_alignment);
 
-                Self::render_label(buf, label, label_area, Alignment::Center);
-            }
+        for (i, label) in labels[1..labels.len() - 1].iter().enumerate() {
+            // We add 1 to x (and width-1 below) to leave at least one space before each intermediate labels
+            let x = graph_area.left() + (i + 1) as u16 * width_between_ticks + 1;
+            let label_area = Rect::new(x, y, width_between_ticks.saturating_sub(1), 1);
+
+            Self::render_label(buf, label, label_area, Alignment::Center);
         }
+
+        let x = graph_area.right() - width_between_ticks;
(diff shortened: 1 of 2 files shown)


src/widgets/chart.rs
```diff
@@ -353,7 +366,7 @@ impl<'a> Chart<'a> {
         layout
     }
 
-    fn max_width_of_labels_left_of_y_axis(&self, area: Rect) -> u16 {
+    fn max_width_of_labels_left_of_y_axis(&self, area: Rect, has_y_axis: bool) -> u16 {
```
fdehau, line 369: is there an advantage of having `has_y_axis` has a parameter instead of the doing `self.y_axis_labels.is_some()` within the method ?

src/widgets/chart.rs
```diff
@@ -385,28 +408,77 @@ impl<'a> Chart<'a> {
...
+
         for (i, label) in labels.iter().enumerate() {
-            let label_width = label.width() as u16;
-            let label_width = if i == 0 {
-                // the first label is put between the left border of the chart and the y axis.
-                graph_area
-                    .left()
-                    .saturating_sub(chart_area.left())
-                    .min(label_width)
+            if i == 0 {
```
fdehau, line 415: we return early if labels_len < 2 therefore wouldn't it be more clear to unroll this loop:
```rust
// first label

// labels[1..labels.len() - 1]

// last label
```

src/widgets/chart.rs
```diff
@@ -362,7 +375,17 @@ impl<'a> Chart<'a> {
             .unwrap_or_default();
         if let Some(ref x_labels) = self.x_axis.labels {
             if !x_labels.is_empty() {
-                max_width = max(max_width, x_labels[0].content.width() as u16);
+                let first_label_width = x_labels[0].content.width() as u16;
```
fdehau, line 378:
```rust
if let Some(first_label_width) = x_labels.get(0).map(|l| l.content.width() as u16) {
  //
}
```
maybe ? to avoid the empty test ?

tests/widgets_chart.rs
```diff
@@ -13,6 +14,22 @@ fn create_labels<'a>(labels: &'a [&'a str]) -> Vec<Span<'a>> {
     labels.iter().map(|l| Span::from(*l)).collect()
 }
 
+fn axis_test_case<S>(width: u16, height: u16, x_axis: Axis, y_axis: Axis, lines: Vec<S>)
```
fdehau, line 17: I wonder if it really makes sense to have a dedicated function for something that is only used in few places for now. Don't you think it would be clearer and more flexible to have each test have it inline ?

fdehau (commented):
Thanks a lot of taking the time to look into this, it looks great. I have left some comments but the PR is already in a very good state — the screenshots and tests are really appreciated 🙇‍♂️. Let me know what you think.

src/widgets/chart.rs
theogilbert: You're right, the loop becomes much clearer once we exclude the first and last label cases from it.

tests/widgets_chart.rs
theogilbert: I felt that this "technical" part of the test case did not add any value in understanding what the tests did. Also, as this code was duplicated across multiple tests it felt right to extract it to a function.

But yes I suppose that this change could hinder future modifications of the tests. If you feel that this function is not worth the trouble, I can inline it back in the tests.

src/widgets/chart.rs
theogilbert: It might just be me, but I feel like too much is going on in this single line.
How about merging the two nested conditions ?

```rust
if let Some(first_x_label) = self.x_axis.labels.as_ref().and_then(|labels| labels.get(0)) {
    let first_label_width = first_x_label.content.width() as u16; 
    //
}
```

src/widgets/chart.rs
theogilbert: The fact that the Y axis is drawn only if it has labels seemed like a piece of logic that would not be obvious in this method.
It would also duplicate the code already present in `Chart::layout()`.

To avoid passing this paramter, another solution that I might propose is to encapsulate this logic in a new method `Chart::should_y_axis_be_rendered(&self)`, referred by both this method and `Chart::layout()`.

fdehau: That makes sense, thanks for the explanation. The new method might indeed be a nice addition but this is not blocking this PR.

fdehau:
Thanks a lot for the latest fixes and for this nice contribution 🔥

PR #593, opened 2022-02-26
adds panic hook example

wookietreiber:
## Description

Without a terminal-resetting panic hook there are two main problems when an application panics:

1.  The report of the panic is distorted because the terminal has not properly left the alternate screen and is still in raw mode.
2.  The terminal needs to be manually reset with the `reset` command.

To avoid this, the standard panic hook can be extended to first reset the terminal. That's what this new example showcases.

## Testing guidelines

```
cargo run --example panic
```

## Checklist

* ~~I have added relevant tests.~~ (shouldn't be required with an example)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,6 +65,10 @@ name = "list"
 required-features = ["crossterm"]
 
 [[example]]
+name = "panic"
+required-features = ["crossterm"]
+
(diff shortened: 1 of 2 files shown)


fdehau:
Thanks a lot 🙇‍♂️

PR #597, opened 2022-03-08
docs: add projects of EdJoPaTo

EdJoPaTo:
## Description

add projects of EdJoPaTo to README.

Personally I think the list should be sorted alphabetically or something like that so each project looks the same instead of "the first are better somehow". But that is not in the scope of this PR.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ You can run all examples by running `cargo make run-examples` (require
 
 ### Third-party widgets
 
+* [tree-widget](https://github.com/EdJoPaTo/tui-rs-tree-widget)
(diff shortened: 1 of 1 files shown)


wcampbell0x2a:
Cool, you should add an example image to your README.md

EdJoPaTo:
my mqttui tool was added in 90d8cb65261247d87fe5f096c32635c1b2049d9c so only my tree widget is added in this Pull Request.

fdehau:
Done in https://github.com/fdehau/tui-rs/pull/659

PR #635, opened 2022-06-23
add helix to the list of apps using tui-rs

NChechulin:
## Description
Just a documentation update: add helix editor to the list of apps using tui-rs

## Testing guidelines
No testing needed

## Checklist


archseer:
Note that we forked and removed most of the implementation though. I mostly use the `buffer.rs` implementation that does efficient render diffing

fdehau:
👋 This is a very nice project. If you are not using `tui` directly, I don't think it make sense to link to it from the README. I'm using this list as way for people to discover projects that might help them develop their own using `tui`. I'm glad that some of the code was still useful though 😅.

PR #644, opened 2022-07-17
Use `SmolStr` to avoid bunch of heap allocations for `Cell`s

### LINKED ISSUE #627: Avoid string allocation on Cell
## Problem
I admit tui-rs isn't supposed to be extra fast. Yet. `Cell::symbol` is a fully formed heap-allocated `String`, which seems like a big waste?

## Solution
Could it just be a `char`?

## Alternatives
If it has to be a string , then it could be `&'a str` at least. The user could create a single string and make the cells see sub-slices of it.

## Additional context
Related to #535

rhysd:
Fixes #627

## Description

To render a `Buffer` instance, `height * length` of `String`s are allocated on heap because one `String` is allocated per one `Cell`. For example, my terminal requires `$LINES * $COLUMNS = 82 * 318 = 26076` of `String` allocations on heap to render an application with entire terminal screen.

Recently I found an awesome crate [`smol_str`](https://github.com/rust-analyzer/smol_str) maintained by rust-analyzer team. This crate provides a string type `SmolStr` with small string optimization (SSO). Its size is always the same as `String`. So size of `Cell` does not change. And it can contain a string without heap allocation up to 22 bytes. I believe almost all unicode grapheme clusters fit to the size. So almost all heap allocations around `Cell` can be removed by simply replacing `String` with `SmolStr`.

This PR replaces the `symbol` field of `Cell`. Since the field is public, this change will lead to major version bump (v0.19.0?). But it is not so big problem because `Cell` is very implementation details and typically TUI application authors don't access it directly.

Thanks to carefully designed `SmolStr`'s APIs, just replacing `String` with `SmolStr` almost worked and changes of this PR are very small. The difference between `SmolStr` and `String` is immutable or not. `symbol` field is now newly created instead of clearing its contents.

## Testing guidelines
- I ran all unit tests and confirmed they passed
- I ran all examples and checked they were working fine manually

## Checklist

* [x] I have added relevant tests. (I believe current test is sufficient)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ unicode-width = "0.1"
 termion = { version = "1.5", optional = true }
 crossterm = { version = "0.23", optional = true }
 serde = { version = "1", optional = true, features = ["derive"]}
+smol_str = "0.1"
 
 [dev-dependencies]
 rand = "0.8"
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -3,6 +3,7 @@ use crate::{
     style::{Color, Modifier, Style},
     text::{Span, Spans},
 };
+use smol_str::SmolStr;
 use std::cmp::min;
 use unicode_segmentation::UnicodeSegmentation;
 use unicode_width::UnicodeWidthStr;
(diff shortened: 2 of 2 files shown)


rhysd:
Let me decline this PR because

- `smol_str` only supports the latest stable Rust
- I've seen some performance regressions in my [tui-textarea](https://github.com/rhysd/tui-textarea) micro benchmarks

PR #653, opened 2022-08-12
Feature: allow word-break in wrapped paragraphs.

chldu2000:
## Description
This PR will allow word-break in wrapped paragraphs.

## Testing guidelines

Create a `Paragraph`  in a `Block`, with the `break_word` value in its `Wrap` set to `true`.

Such as :

```rust
const SAMPLE_STRING: &str = "The library is based on the principle of immediate rendering with \
     intermediate buffers. This means that at each new frame you should build all widgets that are \
     supposed to be part of the UI. While providing a great flexibility for rich and \
     interactive UI, this may introduce overhead for highly dynamic content.";

fn ui<B: Backend>(frame: &mut Frame<B>) {
    let size = Rect::new(0, 0, 20, 10);
    let text = vec![Spans::from(SAMPLE_STRING)];
    let paragraph = Paragraph::new(text)
        .block(Block::default().borders(Borders::ALL))
        .alignment(Alignment::Left)
        .wrap(Wrap {
            trim: true,
            break_words: true,
        });
    frame.render_widget(paragraph, size);
}
```

If set `break_words` as `true`, the result will be like:

```shell
┌──────────────────┐
│The library is bas│
│ed on the principl│
│e of immediate ren│
│dering with interm│
│ediate buffers. Th│
│is means that at e│
│ach new frame you │
│should build all w│
└──────────────────┘
```

If set `break_words` as `false`, the result will be like:

```shell
┌──────────────────┐
│The library is    │
│based on the      │
│principle of      │
│immediate         │
│rendering with    │
│intermediate      │
│buffers. This     │
│means that at each│
└──────────────────┘
```

## Checklist

### DIFF
--- a/examples/demo/ui.rs
+++ b/examples/demo/ui.rs
@@ -293,7 +293,10 @@ where
             .fg(Color::Magenta)
             .add_modifier(Modifier::BOLD),
     ));
-    let paragraph = Paragraph::new(text).block(block).wrap(Wrap { trim: true });
+    let paragraph = Paragraph::new(text).block(block).wrap(Wrap {
+        trim: true,
+        break_words: false,
+    });
     f.render_widget(paragraph, area);
 }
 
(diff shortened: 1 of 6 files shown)


TBS1996:
This is great, I'm really missing this feature

PR #706, opened 2023-05-18
Feature: Ability to mark terminal dirty and force a full repaint on next draw call

karcsesz:
## Description
Adds some new functions to `Cell`, `Buffer`, and `Terminal` that when called marks a given object dirty, meaning their current state on the terminal is unknown, so they should be repainted the next time `draw()` is called.

Uses a given cell's `.symbol` being empty as a "dirty" marker. Not sure how proper this is for an implementation, but it seems to work well.

## Testing guidelines
I've added some tests to `tests/terminal.rs`

## Checklist

* [x] I have read the contributing guidelines.
* [x] I have documented all new additions. I think.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -65,6 +65,11 @@ impl Cell {
         self.bg = Color::Reset;
         self.modifier = Modifier::empty();
     }
+    /// Marks cell dirty so it has to be redrawn
+    pub fn mark_dirty(&mut self) {
+        self.reset();
+        self.symbol.clear();
+    }
 }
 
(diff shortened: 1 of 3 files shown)


orhun:
Hey, [ratatui](https://github.com/tui-rs-revival/ratatui) is the actively maintained fork of `tui-rs`. If you are interested, feel free to submit this PR to there as well. Cheers!

karcsesz:
Thanks @orhun! I've ported my project over, and will see if there's a more elegant way to achieve this functionality. It is a bit of an ugly bodge.

